PR #1, opened 2018-09-28
Pandas dataframes returned by the query API are now singleton cells

ACscooter:
When a property contains multiple values, this gets expanded into multiple rows in the Pandas dataframe.

### DIFF
--- a/datacommons/test/datacommons_test.py
+++ b/datacommons/test/datacommons_test.py
@@ -66,7 +66,7 @@ class AppTest(unittest.TestCase):
     }, {
         'rows': [{
             'cells': [{
-                'value': ['Oceania']
+                'value': ['Oceania', 'North America']
             }, {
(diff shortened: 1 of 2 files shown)


datacommons/test/datacommons_test.py
```diff
@@ -74,8 +74,8 @@ def test_query(self):
         'header': ['name', 'area_sq_mi']
     })
     expected_df = pd.DataFrame({
-        'name': [['Oceania']],
-        'area_sq_mi': [['3292000']]
+        'name': ['Oceania'],
+        'area_sq_mi': ['3292000']
     })
```
pradh: Could you pls add repeated property to validate those become multiple rows.

ACscooter: Fixed f74cdd6 !

PR #14, opened 2018-10-15
Add get_contained_places api

shifucun:


datacommons/datacommons.py
```diff
@@ -452,6 +454,52 @@ def get_states(self, country, new_col_name, max_rows=100):
 
     return pd.concat([type_row, dcid_column], ignore_index=True)
 
+
+  def get_contained_places(self,
```
ACscooter: Were we originally to have arguments `place` (dcid), `contained_in_place_type` instead of `place_type`, `place_name`, and `contained_place_type`?

datacommons/datacommons.py
```diff
@@ -452,6 +454,56 @@ def get_states(self, country, new_col_name, max_rows=100):
...
+    query_result = self.query(type_query)
+    assert query_result['type'].count() == 1, (
+        'Type of the container dcid not found')
+    container_type = query_result['type'][0]
+
+    # Sanity check the type information.
+    place_type_ind = _PLACES.index(place_type)
+    container_type_ind = _PLACES.index(container_type)
+    assert container_type_ind > place_type_ind, (
+        'Asked place should be of lower level type of the container')
```
pradh: nit: 'Requested place should be of a lower level type than the container'

shifucun: Updated

PR #17, opened 2018-10-22
Add save_dataframe api

shifucun:


datacommons/datacommons.py
```diff
@@ -397,6 +398,28 @@ def get_observations(self,
         'Observation',
         max_rows=max_rows)
 
+  # -------------------------- CACHING FUNCTIONS --------------------------
+
+  def save_dataframe(self, pd_dataframe, file_name):
+    """Saves pandas dataframe for later retrieving.
```
pradh: Since this is user-facing. How about adding some comments to say this is saved within a user's namespace and is not currently accessible to any other user?

rvguha:
How do we let a user make their data frame available to all users? Should a
data frame be a Google doc/spreadsheet?

shifucun:
I will extend this function to return a global file name so others can call read_dataframe function to load the dataframe

PR #18, opened 2018-10-23
 Added documentation for area, timezone, freebaseId, geonamesId, and wikidataId.

vholland:


danbri:
LGTM although

>     <div typeof="rdf:Property" resource="https://schema.datacommons.org/timezone">
>       <span property="rdfs:label">timezone</span>
 >     <span property="rdfs:comment">The current timezone for the location.</span>

... does this mean that e.g. UK would have "BST" as true value of this property only half of the year? /cc @vholland @rvguha

vholland:
I debated how to describe this without introducing Role.

vholland:
There are a significant number of places (including at the US city level) that change timezones every six months.  (There is even a nascent movement to have Massachusetts switch to Atlantic time and then opt out of daylight savings.)

With that said, I am willing to remove "current" and let people interpret it as they will.

danbri:
It would be interesting at schema.org level to annotate the volatility of property values (and to mine wikidata for similar info). For now let's remove "current". Everything is current.

vholland:
This is the second conversation I've had today that boils down to "What is time?" I think I need more coffee.

Meanwhile, implemented in PR #21

PR #19, opened 2018-10-23
Add read_dataframe api.

shifucun:


datacommons/datacommons.py
```diff
@@ -400,9 +400,34 @@ def get_observations(self,
 
   # -------------------------- CACHING FUNCTIONS --------------------------
 
+  def read_dataframe(self, file_name):
+    """Read pandas dataframe from caching.
```
pradh: Read a previously saved pandas dataframe.

pradh: cache

datacommons/datacommons.py
```diff
@@ -400,9 +400,34 @@ def get_observations(self,
...
+
+      Use can only read previously saved data file with the same authentication
```
pradh: nit: User

datacommons/examples/simple_query.py
```diff
@@ -45,7 +45,10 @@ def main():
   with pd.option_context('display.width', 400, 'display.max_rows', 100):
     print(df)
 
-  dc.save_dataframe(df, 'simple_query_result')
+  dc.save_dataframe(df, 'test_df')
+  saved_pd = dc.read_dataframe('test_df')
+  with pd.option_context('display.width', 400, 'display.max_rows', 100):
+    print(saved_pd)
```
pradh: Can you compare the original and saved/restored dataframes?

datacommons/datacommons.py
shifucun: Updated

datacommons/examples/simple_query.py
shifucun: Added comparison

PR #30, opened 2018-12-04
Use print() function in both Python 2 and Python 3

cclauss:
__print()__ is a function in Python 3.


Spaceenter:
Thanks - closing this PR as we are doing a major code rewrite which will take into account this.

PR #42, opened 2019-03-23
Fixed simple_query example to have double quotes instead of single quotes

ACscooter:


datacommons/__init__.py
```diff
@@ -14,3 +14,5 @@
 """Data Commons module."""
 
 from .datacommons import Client
```
shifucun: Remove?

ACscooter: Woops, fixed!

PR #45, opened 2019-04-30
docstrings: typos, consistency, proposed improvement

tjann:
please review logic in the larger docstring change proposal (line 328)

### DIFF
--- a/datacommons/datacommons.py
+++ b/datacommons/datacommons.py
@@ -215,7 +215,7 @@ class Client(object):
     Args:
       col_name: Column name for the returned column.
       instance_type: String of the instance type.
-      max_rows: Max number of returend rows.
+      max_rows: Max number of returned rows.
 
     Returns:
       A pandas.DataFrame with instance dcids.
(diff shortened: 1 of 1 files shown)


datacommons/datacommons.py
```diff
@@ -325,13 +325,14 @@ def get_observations(self,
                        max_rows=100):
     """Create a new column with values for an observation of the given property.
 
-    The existing pandas dataframe should include a column containing entity IDs
-    for a certain schema.org type. This function populates a new column with
-    property values for the entities.
+    The existing pandas dataframe should include a column containing population
+    dcids. This function populates a new column with observations of the
+    populations' measured property. A column containing geo ids of type City
+    can be used instead of population dcids.
```
pradh: While you're here, probably worth finding out whether the 'City' limitation is really necessary. (It could just be that at some point only City had directly attached observations.)

tjann: At least on dataCommons.org, looks like some weather observations are attached to cities ("Observation of temperature" https://browser.datacommons.org/kg?dcid=geoId/4261000). Assuming dataCommons.org is up to date, the 'City' qualification seems to be necessary.

tjann: Looks like some of the EPA observations are also attached to cities.

pradh: Oh, I meant to say, can it be *any* entity type rather than just City, like how get_populations() allows.

IOW, is the assert below still required... but that's a question to Bo or Antares.
    assert seed_col_type == 'StatisticalPopulation' or seed_col_type == 'City', (
        'Parent entity should be StatisticalPopulation' or 'City')

PR #57, opened 2019-07-29
Implemented API revisions

ACscooter:


datacommons/utils.py
```diff
@@ -48,84 +50,49 @@
...
-    return pd_frame.dropna(subset=col_names)
-  return process
+def clean_frame(pd_frame):
+  """ A convenience function that cleans a pandas DataFrame.
 
-def delete_column(*cols):
-  """ Returns a function that deletes the given column from a frame.
+  The following operations are performed:
+  - Columns containing numerical types are converted to floats.
+  - Rows with empty values are dropped.
```
shifucun: We might not want to do this but leave it to user. As this has side effects of removing arbitrary rows..

datacommons/utils.py
```diff
@@ -141,3 +108,55 @@ def format_response(response, compress=False):
...
+
+  # Ensure all dcids in must_exist have some entry in results.
+  for dcid in must_exist:
+    results[dcid]
+  return dict(results)
+
+def convert_dcids_type(dcids):
+  """ Amends dcids list type and creates the approprate request dcids list. """
+  # Format the dcids list.
+  if isinstance(dcids, str):
```
shifucun: Might not want the dcids here to be string or only list or pandas series. So this function could be just:

list(dcids)

datacommons/examples/places_basic.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+
+import datacommons as dc
+
+def print_header(label):
+  print('\n' + '-'*80)
+  print(label)
+  print('-'*80 + '\n')
+
+def main():
```
shifucun: I feel we might want to have the basic and pandas as one example. Overall, we might not want to emphasis too much on the two options. We will mention the capability but not making it feel too overwhelming (too many options for user)

datacommons/places.py
```diff
@@ -12,44 +12,41 @@
...
+  url = utils._API_ROOT + utils._API_ENDPOINTS['get_place_in']
+  res = requests.post(url, json={
+    'dcids': req_dcids,
+    'place_type': place_type,
+    'reload': reload,
+  })
+  payload = utils.format_response(res)
+
+  # Create the results and format it appropriately
+  result = utils.format_expand_payload(payload, 'place', must_exist=dcids)
```
shifucun: what is must_exists used for?

datacommons/populations.py
```diff
@@ -12,180 +12,97 @@
...
-
-  Raises:
-    ValueError: when input argument is not valid.
-  """
-  if seed_col_name not in self._dataframe:
-    raise ValueError('{} is not a valid seed column.'.format(seed_col_name))
-  if new_col_name in self._dataframe:
-    raise ValueError('{} is already a column name.'.format(new_col_name))
+def get_populations(dcids, population_type, pv_map, reload=False):
+  """ Returns a list of populations associated with the given dcids.
```
shifucun: this could be optional map for populations with no pvs

datacommons/core.py
```diff
@@ -0,0 +1,130 @@
...
+  payload = utils.format_response(res)
+
+  # Return the results based on the orientation
+  if outgoing and 'outArcs' in payload:
+    return payload['outArcs']
+  elif not outgoing and 'inArcs' in payload:
+    return payload['inArcs']
+  return []
+
+def get_property_values(dcids, prop, outgoing=True, value_type=None, reload=False, limit=utils._MAX_LIMIT):
```
Spaceenter: line size <=80

ditto for the rest of the PR.

datacommons/core.py
```diff
@@ -0,0 +1,130 @@
...
+
+def get_triples(dcid, reload=False, limit=utils._MAX_LIMIT):
+  """ Returns a list of triples where the dcid is either a subject or object.
+
+  The return value is a list of tuples (s, p, o) where s denotes the subject
+  entity, p the property, and o the object.
+
+  Args:
+    dcid: The node to get triples for.
+    reload: Whether or not to send the query to cache.
```
Spaceenter: This is confusing. Reading this feels like "reload=true" sends query to cache, which is not true.

datacommons/examples/core_basic.py
```diff
@@ -0,0 +1,53 @@
...
+  print_header('Property Values for "containedInPlace" of Santa Clara County')
+  prop_vals = dc.get_property_values(dcid, 'containedInPlace', outgoing=False, value_type='City', reload=True)
+  print('> Cities contained in {}'.format(dcid))
+  if dcid in prop_vals:
+    for city_dcid in prop_vals[dcid]:
+      print('  - {}'.format(city_dcid))
+
+  # Print the first 10 triples associated with Santa Clara County
+  print_header('Triples for "containedInPlace" of Santa Clara County')
+  triples = dc.get_triples(dcid, reload=True)
```
Spaceenter: No need to set "reload=True"?

datacommons/examples/populations_pandas.py
```diff
@@ -0,0 +1,56 @@
...
+
+"""
+
+import datacommons as dc
+import pandas as pd
+
+pd.options.display.max_rows = 10
+pd.options.display.max_colwidth = 30
+
+def print_header(label):
```
Spaceenter: Seems like we have print_header in every example py file. Could it be extracted out to a util lib?

datacommons/places.py
```diff
@@ -12,44 +12,41 @@
...
+  If the dcids field is a list, then the return value is a dictionary mapping
+  dcid to the list of values associated with the given property.
+
+  If the dcids field is a Pandas Series, then the return value is a Series where
+  the i-th cell is the list of values associated with the given property for the
+  i-th dcid.
+
+  Args:
+    dcids: List of dcids to get contained in places of.
+    place_type: The type of place to query for.
```
Spaceenter: s/place/places

datacommons/utils.py
ACscooter: Good point, removed

datacommons/populations.py
ACscooter: Good point, changed

datacommons/core.py
ACscooter: Agreed, fixed to say "A flag that sends the query without hitting cache when set."

datacommons/examples/populations_pandas.py
ACscooter: Yeah this sounds good.

datacommons/places.py
ACscooter: Fixed

datacommons/utils.py
ACscooter: Discussed offline, decided to keep for now.

datacommons/populations.py
tjann: Ok for now, but left a comment in notebook about "pv_map" not being an informative var name. Can change later after discussion

datacommons/core.py
```diff
@@ -0,0 +1,130 @@
...
+from collections import defaultdict
+
+import pandas as pd
+
+import datacommons.utils as utils
+import requests
+
+# ----------------------------- WRAPPER FUNCTIONS -----------------------------
+
+def get_property_labels(dcid, outgoing=True, reload=False):
```
tjann: I guess the assumption is that a node will never have too many properties. For now that seems more than reasonable, but stands out because almost all other methods have a limit on return length.

datacommons/examples/places_basic.py
ACscooter: Yeah I agree, I've merged the examples together into a single file.

datacommons/populations.py
```diff
@@ -12,180 +12,97 @@
...
-  # Perform the query and merge the results
-  new_frame = Frame(datalog_query=query, labels=labels, type_hint=type_hint, rows=rows)
-  self.merge(new_frame)
-
-def get_observations(self,
-                     seed_col_name,
-                     new_col_name,
-                     observation_date,
-                     measured_property,
-                     stats_type=None,
```
tjann: Nit: can you swap measured_property and obsevation_date, then move stats_type down? That way we have logical flow of observation_date followed by observation_period

datacommons/places.py
ACscooter: Ah it ensures that certain dcids are returned in the payload. I'll remove this once the right output is returned by the mixer.

datacommons/core.py
ACscooter: Removed reload from the API

datacommons/examples/core_basic.py
ACscooter: Removed reload from the API

PR #61, opened 2019-08-02
Added unit tests

ACscooter:
Additional changes include
- Clarified comments and documentation
- Updated client API parameters to match latest mixer release


datacommons/core.py
```diff
@@ -95,33 +93,36 @@ def get_property_values(dcids,
...
-  The return value is a list of tuples (s, p, o) where s denotes the subject
-  entity, p the property, and o the object.
+  The return value is a dictionary mapping given dcids to list of triples. The
+  triples are repsented as 3-tuples (s, p, o) where "s" denotes the subject, "p"
+  the property, and "o" the object.
 
   Args:
-    dcid: A list of dcids to get triples for.
+    dcids: A list of dcids to get triples for.
     limit: The maximum number of triples to get for each combination of property
```
tjann: can say property label to be more clear

datacommons/__init__.py
```diff
@@ -13,13 +13,13 @@
 # limitations under the License.
 """Data Commons module."""
 
-from __future__ import absolute_import
-from __future__ import division
```
tjann: no matter how I refresh, I still see this as deleted. will come back to it but can you check that this was put back?

ACscooter: Added back

datacommons/core.py
ACscooter: Will clarify with a section on what a triple is.

PR #69, opened 2019-08-07
Check in R API Client code

tjann:


tjann:
Should I indicate that this is beta within the R folder's README? Or will this be general?

PR #71, opened 2019-08-07
Fix typo in places.py.

lukesmurray:


lukesmurray:
@googlebot I signed it!

PR #82, opened 2019-08-09
Restructured documentation to have ToCs

ACscooter:


docs/source/index.rst
```diff
@@ -22,7 +22,6 @@ to use and understand!
 
 - Send an issue request to the
   `Data Commons Issues Page <https://github.com/google/datacommons/issues>`_.
-  When creating an issue please mark the issue using the **api feedback** label!
```
beets: why don't we want our issues tagged anymore?

ACscooter: Turns out non-contributors can't assign labels to their github issues. But, Github does allow issue templating. I've made a template for python API issues @ 00cb027f

Issue #67, opened 2019-08-07
Add uninstall or upgrade instructions


panesargoog:
example:

pip install --upgrade git+https://github.com/google/datacommons.git@dev

(also instructions on pip, python versions required)

lukesmurray:
For upgrading its common for projects to use the `-U` flag in their install instructions.

```
pip install -U git+https://github.com/google/datacommons.git@dev
```

PR #86, opened 2019-08-12
Added iPynb notebooks and fixed issue #67

ACscooter:
Commit a62d777 fixes issue #67


ACscooter:
We can save directly from colab.

image

beets:
> We can save directly from colab.
> 
> image

Nice, could you add a note to that effect in the README?

PR #93, opened 2019-08-17
Add get_pop_obs api

shifucun:


datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
   if isinstance(dcids, pd.Series):
     return pd.Series([result[dcid] for dcid in dcids])
   return result
+
+
+def get_pop_obs(dcid):
+  """ Returns :obj:`Place`'s population and observations.
```
ACscooter: Returns all :obj:\`StatisticalPopulation\`'s and :obj:\`Observation\`'s of a :obj:\`Place\`

datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
...
+
+  Args:
+    dcids (:obj:`str`): Dcid of the place.
```
ACscooter: dcid instead of "dcids"

datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
...
+
+  Returns:
+    An dictionary of population and observation related to this place.
```
ACscooter: "A :obj:\`dict\` of :obj:\`StatisticalPopulation\`'s and :obj:\`Observation\`'s that are associated to the place identified by the given :code:\`dcid\`. The given dcid is the :obj:\`location\` of the returned :obj:\`StatisticalPopulation\`'s, which are the :obj:\`observedNode\`'s of the returned :obj:\`Observation\`'s. See example below for more detail about how the returned :obj:\`dict\` is structured."

We may want to link to the browser nodes for StatisticalPopulation, Observation and location.

datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
...
+
+  Raises:
+    ValueError: If the payload returned by the Data Commons REST API is
+      malformed.
+
+  Examples:
+    We would like to get population and observation of Mountain View
```
ACscooter: :obj:\`StatisticalPopulation\`'s and :obj:\`Observations\`'s of \`Mountain View <https://browser.datacommons.org/kg?dcid=geoId/0649670>\`_.

datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
...
+    `Mountain View <https://browser.datacommons.org/kg?dcid=geoId/0649670>`_.
+
+    >>> get_pop_obs("geoId/0649670")
+    {
+      'name': 'Mountain View',
+      'placeType': 'City',
+      'populations': {
+        'dc/p/013ldrstf6lnf': {
+          'numConstraints': 6,
```
ACscooter: Can we order the keys as follows: popType, numConstraints, propertyValues, observations?

datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
   if isinstance(dcids, pd.Series):
     return pd.Series([result[dcid] for dcid in dcids])
   return result
+
+
+def get_pop_obs(dcid):
```
ACscooter: Should we move this to the Populations module? The language of get_pop_obs documentation matches more get_populations and get_observations, than it does get_places_in.

datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
...
+      'placeType': 'City',
+      'populations': {
+        'dc/p/013ldrstf6lnf': {
+          'numConstraints': 6,
+          'observations': [
+            {
+              'marginOfError': 119,
+              'measuredProp': 'count',
+              'measuredValue': 225,
+              'measurementMethod': 'CenusACS5yrSurvey',
```
ACscooter: Can we choose BLS as the example population? We should try to avoid any observation that has measurementMethod "Cenus"ACS5YrSurvey until that's fixed in the graph.

datacommons/places.py
```diff
@@ -92,3 +92,59 @@ def get_places_in(dcids, place_type):
...
+            'age': 'Years16Onwards',
+            'gender': 'Male',
+            'income': 'USDollar30000To34999',
+            'incomeStatus': 'WithIncome',
+            'race': 'USC_HispanicOrLatinoRace',
+            'workExperience': 'USC_NotWorkedFullTime'
+          }
+        }
+      }
+    }
```
ACscooter: At the bottom, shall we add/modify this snippet of text?

```
Notice that the return value is a multi-level :obj:`dict`. The top level contains the following keys.

- :code:`name` and  :code:`placeType` provides the name and type of the :obj:`Place` identified by the given :code:`dcid`.
- :code:`populations` maps to a :obj:`dict` containing all :obj:`StatisticalPopulation`'s that have the given :code:`dcid` as its :obj:`location`.

The :code:`populations` dictionary is keyed by the dcid of each :obj:`StatisticalPopulation`. The mapped dictionary contains the following keys.

- :code:`popType` which gives the population type of the :obj:`StatisticalPopulation` identified by the key.
- :code:`numConstraints` which gives the number of constraining properties defined for the identified :obj:`StatisticalPopulation`.
- :code:`propertyValues` which gives a :obj:`dict` mapping a constraining property to its value for the identified :obj:`StatisticalPopulation`. 
- :code:`observations` which gives a list of all :obj:`Observation`'s that have the identified :obj:`StatisticalPopulation` as their :obj:`observedNode`. Each :obj:`Observation` is represented by a :code:`dict` that have the keys: <INSERT AND DOCUMENT ALL KEYS HERE>
```

datacommons/examples/places.py
```diff
@@ -56,6 +57,11 @@ def main():
   pd_frame = dc.flatten_frame(pd_frame)
   print(pd_frame)
 
+  # Get all population and observation data of Mountain View.
```
ACscooter: Maybe move to datacommons/examples/populations.py. See comment below about moving get_pop_obs to populations submodule.

datacommons/places.py
shifucun: This is very good!

datacommons/places.py
ACscooter: Ah okay that's fine. This raises a couple questions for me actually.

- I think the distinction between get_place_obs and get_pop_obs is a bit confusing now. I would have assumed that observations linked directly to place would be returned by get_place_obs.
- Also how is the structure of the return value if the observation is linked directly to the place dcid (in particular, how do you differentiate Observations linked by "location" vs. "childhoodLocation"? 
- Is it possible to separate get_pop_obs into two functions with one returning population observation pairs, and the other returning observations linked directly to the Place?

datacommons/places.py
```diff
@@ -29,7 +29,7 @@
 
 
 def get_places_in(dcids, place_type):
-  """ Returns :obj:`Place`'s contained in :code:`dcids` of type `place_type`.
+  """ Returns :obj:`Place`s contained in :code:`dcids` of type `place_type`.
```
ACscooter: :code:\`place_type\`.

datacommons/places.py
```diff
@@ -95,24 +95,24 @@ def get_places_in(dcids, place_type):
...
   Args:
     dcid (:obj:`str`): Dcid of the place.
 
   Returns:
-    "A :obj:`dict` of :obj:`StatisticalPopulation`'s and :obj:`Observation`'s that are associated to the place
-    identified by the given :code:`dcid`. The given dcid is the :obj:`location` of the returned :obj:`StatisticalPopulation`'s,
-    which are the :obj:`observedNode`'s of the returned :obj:`Observation`'s.
-    See example below for more detail about how the returned :obj:`dict` is structured."
+    A :obj:`dict` of :obj:`StatisticalPopulation` and :obj:`Observation` that are associated to the place
+    identified by the given :code:`dcid`. The given dcid is the :obj:`location` of the returned :obj:`StatisticalPopulation`,
```
ACscooter: 80 character cap.

datacommons/places.py
```diff
@@ -155,14 +155,23 @@ def get_pop_obs(dcid):
     Notice that the return value is a multi-level :obj:`dict`. The top level contains the following keys.
 
     - :code:`name` and :code:`placeType` provides the name and type of the :obj:`Place` identified by the given :code:`dcid`.
-    - :code:`populations` maps to a :obj:`dict` containing all :obj:`StatisticalPopulation`'s that have the given :code:`dcid` as its :obj:`location`.
+    - :code:`populations` maps to a :obj:`dict` containing all :obj:`StatisticalPopulation` that have the given :code:`dcid` as its :obj:`location`.
```
ACscooter: 80 char cap

datacommons/places.py
```diff
@@ -155,14 +155,23 @@ def get_pop_obs(dcid):
...
 
     The :code:`populations` dictionary is keyed by the dcid of each :obj:`StatisticalPopulation`. The mapped dictionary contains the following keys.
 
     - :code:`popType` which gives the population type of the :obj:`StatisticalPopulation` identified by the key.
     - :code:`numConstraints` which gives the number of constraining properties defined for the identified :obj:`StatisticalPopulation`.
```
ACscooter: 80 char cap and following lines

datacommons/places.py
beets: I'm also wondering if it all needs to be returned in one call. As is, the call seems very browser specific. Also, are these calls place specific? If I understand correctly, we could have statistical population on non-place nodes.

Would we run into limit issues with this call? Or is everything going to be in the cache.

shifucun: > Ah okay that's fine. This raises a couple questions for me actually.
> 
> * I think the distinction between get_place_obs and get_pop_obs is a bit confusing now. I would have assumed that observations linked directly to place would be returned by get_place_obs.
They do have things in common, kinda like get_triples and get_property_values. get_place_obs get observations for a particular population spec, while get_pop_obs get everything. We need to inform users from the comments.

> * Also how is the structure of the return value if the observation is linked directly to the place dcid (in particular, how do you differentiate Observations linked by "location" vs. "childhoodLocation"?
There is an "observations" field to contain all those. Currently there is no way to tell that, this is a bug.

> * Is it possible to separate get_pop_obs into two functions with one returning population observation pairs, and the other returning observations linked directly to the Place?

Might be easier to put them in one, this is like get_triples to get everything. get_place_obs is like get_prop_values.

datacommons/utils.py
```diff
@@ -138,7 +138,7 @@ def clean_frame(pd_frame):
 # ------------------------- INTERNAL HELPER FUNCTIONS -------------------------
 
 
-def _send_request(req_url, req_json={}, compress=False):
+def _send_request(req_url, req_json={}, compress=False, post=True):
```
beets: Oh, interesting -- why are there only GET calls for popops and placekml?

shifucun: get_pop_obs and place_kml only requires a single param (dcid). The other api needs POST as they need to accept multiple parameters that include list, otherwise we can just use GET for those.

datacommons/places.py
shifucun: @beets, that's a good point, we do have SP linked to School etc, that said it seems reasonable to separate out get_pop_obs and get_obs. Also we should introduce "population" and "observation" modules in python.

I suggest we could discuss and do this later as it also involves mixer change.

datacommons/places.py
shifucun: Fixed the 80char issue

datacommons/places.py
shifucun: This is only from cache, so limit is not an issue.

shifucun: get_place_obs gives observation of places, with the population specification fixed.

get_pop_obs actually or (ultimately) get observations for a place, i would see the population here is a by-product. This is like get city in USA, the county is just a by-product.

beets: If this could return non-place results, should this move to the populations module? Unless the mixer currently only returns place results. Seeing the layout of functions across modules now, I think this belongs with populations.. sorry if I misunderstood your comments.

shifucun: Yea, moved to populations module

datacommons/places.py
ACscooter: Thank you :)

datacommons/places.py
ACscooter: Thank you!

datacommons/places.py
shifucun: um, I made it Census now..

datacommons/places.py
beets: thanks!

datacommons/places.py
```diff
@@ -95,82 +96,109 @@ def get_places_in(dcids, place_type):
...
-              'observationPeriod': 'P1M'
+              'measuredValue': 1388,
+              'measurementMethod': 'CenusACS5yrSurvey',
+              'observationDate': '2013'
             }
           ],
         },
-      }
+      },
+      'observations': [
```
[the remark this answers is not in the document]
shifucun: Done

PR #105, opened 2019-11-05
Remove dependency on pandas

shifucun:


datacommons/populations.py
```diff
@@ -88,34 +80,22 @@ def get_populations(dcids, population_type, constraining_properties={}):
...
     'pvs': pv,
   })
 
   # Create the results and format it appropriately
   result = utils._format_expand_payload(
     payload, 'population', must_exist=dcids)
-  if isinstance(dcids, pd.Series):
-    flattened = utils._flatten_results(result, default_value="")
-    return pd.Series([flattened[dcid] for dcid in dcids], index=dcids.index)
+  print(result)
```
ACscooter: Should these print statements be removed?

shifucun: Done

PR #109, opened 2020-01-15
Use the new direction param in GetPropertyValues REST API.

Spaceenter:


datacommons/core.py
```diff
@@ -154,10 +154,14 @@ def get_property_values(dcids,
   """
   # Convert the dcids field and format the request to GetPropertyValue
   dcids = list(dcids)
+  direction = 'in'
```
pradh: nit: Just looks better and we assign direction once when out=True... but feel free to ignore.

if out:
  direction = 'out'
else:
  direction = 'in'

PR #112, opened 2020-01-27
Replace requests with urllib and further clean up pandas

shifucun:


datacommons/examples/core.py
```diff
@@ -49,41 +44,13 @@ def main():
...
-      pd_frame['county'], 'containedInPlace', out=False, value_type='City'))
-  pd_frame = pd_frame.explode('city')
-  print(pd_frame)
-
-  # Get the names for each city.
-  utils._print_header('Get City Names')
-  pd_frame['city_name'] = pd_frame['city'].map(
-    dc.get_property_values(pd_frame['city'], 'name'))
-  pd_frame = pd_frame.explode('city_name')
-  print(pd_frame)
```
pradh: Do we not want to demo these API calls any more?

shifucun: pandas is creating issues when using with Bazel, it can make the entire example not running, and it is the reason to remove them.  Also these are supplemental but not critical for the example, as the API itself does not mention pandas at all.

PR #116, opened 2020-02-05
Make some Python API methods compatible with coming Mixer API changes.

mvashishtha:
We are going to change the places-in, population, and observation mixer API methods
so that they return maps from dcid to lists of dcids, population dcids, and
observation values respectively. Update the corresponding Python API methods so
that they work with both the new and old Mixer API versions.

### DIFF
--- a/datacommons/test/populations_test.py
+++ b/datacommons/test/populations_test.py
@@ -31,8 +31,10 @@ import unittest
 import urllib
 import zlib
 
-
-def request_mock(*args, **kwargs):
+# new_mixer_api specifies whether to use the Mixer API
+# that gives dictionaries instead of lists for places-in,
+# population, and observation functions.
+def request_mock_base(*args, **kwargs):
   """ A mock urlopen request sent in the requests package. """
   # Create the mock response object.
   class MockResponse:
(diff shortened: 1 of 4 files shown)


datacommons/test/populations_test.py
```diff
@@ -68,20 +70,21 @@ def read(self):
...
-          'population': 'dc/p/crgfn8blpvl35'
-        },
-        {
-          'dcid': 'geoId/4805000',
-          'population': 'dc/p/f3q9whmjwbf36'
-        }
-      ])
+      res_json =json.dumps({
+          'geoId/06085': 'dc/p/crgfn8blpvl35',
+          'geoId/4805000': 'dc/p/f3q9whmjwbf36'}) if kwargs['new_mixer_api'] else json.dumps([
```
shifucun: might use a normal if else outside

datacommons/test/populations_test.py
```diff
@@ -103,7 +106,7 @@ def read(self):
     and data['measurement_method'] == 'BLSSeasonallyAdjusted':
     if data['dcids'] == ['dc/p/x6t44d8jd95rd', 'dc/p/lr52m1yr46r44', 'dc/p/fs929fynprzs']:
       # Response returned when querying for multiple valid dcids.
-      res_json = json.dumps([
+      res_json = json.dumps({'dc/p/x6t44d8jd95rd': '18704962.000000', 'dc/p/lr52m1yr46r44': '3075662.000000', 'dc/p/fs929fynprzs': '1973955.000000'}) if kwargs['new_mixer_api'] else json.dumps([
```
shifucun: try to conform to line 80

datacommons/test/populations_test.py
```diff
@@ -120,7 +123,7 @@ def read(self):
       return MockResponse(json.dumps({'payload': res_json}))
     if data['dcids'] == ['dc/p/x6t44d8jd95rd', 'dc/MadDcid']:
       # Response returned when querying for a dcid that does not exist.
-      res_json = json.dumps([
+      res_json = json.dumps({'dc/p/x6t44d8jd95rd' : '18704962.000000'}) if kwargs['new_mixer_api'] else json.dumps([
```
shifucun: same here

PR #124, opened 2020-04-27
Make get_stats more user-friendly. Also help Mixer out by chunking re…

tjann:
…quests.

### DIFF
--- a/datacommons/places.py
+++ b/datacommons/places.py
@@ -72,17 +72,21 @@ def get_places_in(dcids, place_type):
   result = utils._format_expand_payload(payload, 'place', must_exist=dcids)
   return result
   
-def get_stats(dcids, stats_var):
+def get_stats(dcids, stats_var, obs_dates='latest'):
   """ Returns :obj:`TimeSeries` for :code:`dcids` \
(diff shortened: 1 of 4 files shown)


datacommons/test/places_test.py
```diff
@@ -247,6 +247,60 @@ def test_multiple_dcids(self, urlopen):
             }
         })
```
pradh, line 249: Have a test for validating multiple batches? (e.g., override _QUERY_BATCH_SIZE to a small value and expect >1 calls)

datacommons/places.py
```diff
@@ -131,12 +135,35 @@ def get_stats(dcids, stats_var):
...
+      for geo, stats in payload.items():
+        time_series = stats.get('data')
+        if not time_series: continue
+        max_date = max(time_series)
+        max_date_stat = time_series[max_date]
+        time_series.clear()
+        time_series[max_date] = max_date_stat
+        res[geo] = stats
+    elif obs_dates:
+      obs_dates = list(obs_dates)
```
pradh, line 157: nit: if you're only going to look ups, could make it set.

PR #131, opened 2020-07-21
Remove API key requirement and rebrand to Python API (not Python Clie…

tjann:
…nt API).


tjann:
Tested in a notebook-- but fixed the date after.

image

datacommons/utils.py
```diff
@@ -57,27 +58,13 @@
...
-
-  For more details about how to get an API key and provide it to the Python
-  Client API, please visit :ref:`getting_started`
-
-  Args:
-    api_key (:obj:`str`): The api key.
-  """
-  os.environ[_ENV_VAR_API_KEY] = api_key
+  """DEPRECATED FUNCTION--API keys no longer required."""
+  logging.warning('Data Commons has removed the API key requirement. This function will be removed by Dec 1st, 2020.')
```
tjann, line 67: @beets wanted you to review this decision :)

shifucun: Lg. Do we have other docs/codes to remove this as well?

tjann: Yeah need to update docsite... I think that's it. I went through the api-python repo pretty carefully to remove api key refs.

tjann:
Thanks for review! @beets I can submit a patch if you have any other feedback.

datacommons/utils.py
beets: Thanks for calling this out!

I don't think we want to remove this completely. We're allowing you to use it without a key, but if you already have one, we should continue allowing you to set it.

PR #133, opened 2020-07-22
Add mmethod, unit, and observation period to get_stats args.

tjann:


datacommons/places.py
```diff
@@ -138,10 +144,17 @@ def get_stats(dcids, stats_var, obs_dates='latest'):
   batches =  -(-len(dcids) // utils._QUERY_BATCH_SIZE)  # Ceil to get # of batches.
   res = {}
   for i in range(batches):
-    payload = utils._send_request(url, req_json={
+    req_json={
```
pradh, line 147: nit: req_json = {

PR #135, opened 2020-07-27
Let users set API key and pass it along to Mixer.

tjann:


tjann:
Tested first comment with 2 RC Colabs (one get_stats, one Bio) in playground mode on my branch. The work with or without API key.

README.md
```diff
@@ -9,6 +9,15 @@ Once the package is installed, import `datacommons`.
 
     import datacommons as dc
 
+If you would like to provide an API key, enable the Data Commons API for a
+Google Cloud Project and create an API key with Data Commons API access.
+Then, add the following line to your code.
+
+    dc.set_api_key('YOUR-API-KEY')
+
+Data Commons *does not charge* users, but uses the API key for logging and
```
beets, line 18: maybe remove logging from this statement.

README.md
```diff
@@ -9,6 +9,15 @@ Once the package is installed, import `datacommons`.
 
     import datacommons as dc
 
+If you would like to provide an API key, enable the Data Commons API for a
+Google Cloud Project and create an API key with Data Commons API access.
```
beets, line 13: do we want to include a link to the doc page for this?

datacommons/test/set_api_key_test.py
```diff
@@ -0,0 +1,83 @@
...
+class TestQuery(unittest.TestCase):
+  """Unit test for setting the API Key."""
+  @mock.patch('six.moves.urllib.request.urlopen', side_effect=request_mock)
+  def test_no_rows(self, urlopen):
+    """ Handles row-less response. """
+    # Set the API key	
+    dc.set_api_key('TEST-API-KEY')
+
+    # Create a SPARQL query
+    query_string = ('''
```
beets, line 70: no need to change this, but there is probably a simpler call to use here, since the point of the test is setting the key, not testing the query api. i suppose even an empty query should suffice here.

datacommons/test/set_api_key_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def read(self):
+      return self.json_data
+
+  req = args[0]
+  data = json.loads(req.data)
+  # If the API key does not match, then return 403 Forbidden	
+  api_key = req.get_header('X-api-key')
+  if api_key != 'TEST-API-KEY':
+    return urllib.error.HTTPError(None, 403, None, None, None)
```
beets, line 48: this is interesting. so if setting a key, we validate, if not, pass through.

can you add a test to verify the alternate behavior, when a key isn't set? even if it's already tested implicitly elsewhere, would be nice to have an explicit test here.

README.md
tjann: Done

datacommons/test/set_api_key_test.py
tjann: The validation is just to make sure what is set is what we specified. I've changed it to an environment var to test that what we set is what is sent. I also added a foobar before the final set to test this.

Adding a test for no set.

datacommons/test/set_api_key_test.py
tjann: I made it some dummy query (I use it to distinguish the mock return value)

tjann:
PTAL

PR #142, opened 2020-08-20
New get stat functions

tjann:


datacommons/examples/statvar.py
```diff
@@ -0,0 +1,81 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Basic examples for StatisticalVariable-based Data Commons API functions."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import sys
+sys.path.append('../')
```
shifucun, line 21: not sure if this is best practice. Can you check see how people think. I remember we used to run, in the top level folder, python -m datacommons.examples.xxx

datacommons/examples/statvar.py
```diff
@@ -0,0 +1,81 @@
...
+sys.path.append('../../')
+import datacommons as dc
+
+
+def main():
+    # Dcid for Santa Clara County.
```
shifucun, line 27: can you make the entire example parameterized:

data = [{place: xx, stat_var = xxx, xx}, {},{}]
for d in data:
   print(xx)
   print(dc.get())

datacommons/statvar.py
```diff
@@ -0,0 +1,147 @@
...
+      date (:obj:`str`): Optional, the preferred date of observation
+        in ISO 8601 format. If not specified, returns the latest observation.
+      measurement_method (:obj:`str`): Optional, the dcid of the preferred
+        `measurementMethod` value.
+      observation_period (:obj:`str`): Optional, the preferred
+        `observationPeriod` value.
+      unit (:obj:`str`): Optional, the dcid of the preferred `unit` value.
+      scaling_factor (:obj:`int`): Optional, the preferred `scalingFactor` value.
+    Returns:
+      A :obj:`double` the value of :code:`stat_var` for :code:`place`, filtered
```
shifucun, line 72: in python it should be "float" instead of double

datacommons/examples/statvar.py
tjann: Ok, removed and added that to readme.

datacommons/statvar.py
tjann: Thanks, changed

datacommons/examples/statvar.py
tjann: done

datacommons/examples/statvar.py
```diff
@@ -21,57 +21,78 @@
...
+        },
+        {
+            'place': 'geoId/06085',
+            'stat_var': 'UnemploymentRate_Person',
+            'observation_period': "P1Y",
+            'measurement_method': "BLSSeasonallyUnadjusted",
+        },
+        {
+            'place':
+                'nuts/HU22',
```
shifucun, line 57: no need to newline here and below?

datacommons/examples/statvar.py
```diff
@@ -0,0 +1,99 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import datacommons as dc
+
+
+def main():
+    data = [
```
beets, line 24: consider naming this something more specific -- params, etc?

datacommons/examples/statvar.py
```diff
@@ -0,0 +1,99 @@
...
+            'observation_period':
+                "P1Y",
+            'unit':
+                "PurchasingPowerStandard"
+        },
+    ]
+
+    for d in data:
+        print('\n>>> get_stat_value: ',
+              [param for param in d.values() if param])
```
beets, line 75: is the "if param" required? i think it'll be nice to format this as:
>>> get_stat_value(place=geoid/foo, stat_var=statVarFoo)
<<< ...

datacommons/examples/statvar.py
```diff
@@ -0,0 +1,99 @@
+# Copyright 2020 Google Inc.
```
beets, line 1: nit: should this file name be "get_stat.py"?

datacommons/statvar.py
```diff
@@ -0,0 +1,147 @@
+# Copyright 2020 Google Inc.
```
beets, line 1: nit: ditto re: filename. maybe statvars or stat_vars or stats to mirror the other module names?

datacommons/statvar.py
```diff
@@ -0,0 +1,147 @@
...
+
+import json
+import os
+import six.moves.urllib.error
+import six.moves.urllib.request
+
+import datacommons.utils as utils
+
+
+def _send_get_stat_req(url):
```
beets, line 33: this seems mostly identical with _send_request -- does that not work for this module?

tjann: _send_request is expecting a payload keyword, but the REST API does not have that for /stat. It's an open item for discussion.

datacommons/examples/statvar.py
tjann: This is from yapf formatting, will keep since we plan to enforce google pylint later.

datacommons/statvar.py
beets: i wouldn't fork just for that reason. can we add support for non-payload requests?

datacommons/examples/statvar.py
tjann: Done

datacommons/statvar.py
tjann: Ok, I can make it plural

PR #154, opened 2020-09-08
Batch get stat all if more places than default query limit

tjann:
Before code changes:
Screen Shot 2020-09-08 at 10 03 14 AM

After code changes:
Screen Shot 2020-09-08 at 10 04 51 AM


tjann:
Erm, let me try to fix the git history.

PR #155, opened 2020-09-08
Batch get_stat_all calls if more than QUERY_BATCH_SIZE places.

tjann:
Before code changes:

Screen Shot 2020-09-08 at 10 03 14 AM

After code changes:
Screen Shot 2020-09-08 at 10 04 51 AM


datacommons/stat_vars.py
```diff
@@ -203,17 +203,33 @@ def get_stat_all(places, stat_vars):
...
-        raise ValueError('No data in response.')
-
-    # Unnest the REST response for keys that have single-element values.
-    place_statvar_series = collections.defaultdict(dict)
-    for place_dcid, place in res_json['placeData'].items():
-        for stat_var_dcid, stat_var in place['statVarData'].items():
-            place_statvar_series[place_dcid][stat_var_dcid] = stat_var
-    return dict(place_statvar_series)
+    places = list(places)
+    # Ceil hack to get # of batches.
```
beets, line 207: can you describe this hack more?

datacommons/stat_vars.py
```diff
@@ -203,17 +203,33 @@ def get_stat_all(places, stat_vars):
...
+
+        # Unnest the REST response for keys that have single-element values.
+        place_statvar_series = collections.defaultdict(dict)
+        for place_dcid, place in res_json['placeData'].items():
+            for stat_var_dcid, stat_var in place['statVarData'].items():
+                place_statvar_series[place_dcid][stat_var_dcid] = stat_var
+        res.update(dict(place_statvar_series))
+
+    if no_data:
+        raise ValueError('No data in responses.')
```
beets, line 234: This is interesting - so we throw an error even if one out of many batches fail. Is that the right response (perhaps we can add more info in the error)? We should also throw earlier instead of doing this outside the loop.

beets (changes requested):

datacommons/stat_vars.py
```diff
@@ -203,17 +203,33 @@ def get_stat_all(places, stat_vars):
...
-    place_statvar_series = collections.defaultdict(dict)
-    for place_dcid, place in res_json['placeData'].items():
-        for stat_var_dcid, stat_var in place['statVarData'].items():
-            place_statvar_series[place_dcid][stat_var_dcid] = stat_var
-    return dict(place_statvar_series)
+    places = list(places)
+    # Ceil hack to get # of batches.
+    batches = -(-len(places) // utils._QUERY_BATCH_SIZE)
+    res = {}
+    no_data = True
```
shifucun, line 210: it might be better to use has_data here?

datacommons/stat_vars.py
shifucun: Also thinking about the behavior, shall we just return empty result for the entry with no data in stead of throwing error? For example, if someone get stats all for crime for city, county, state, it seems a bit intrusive to throw error because we don't have data for county?

tjann: I think it's actually only returning no_data if placeData was not in any of the responses. no_data is initally set to True, then each time placeData exists (the most top level key of REST response), then no_data is set to False. It is never set to True again. I could optimize it s.t. we only assign False if it is currently True, but the logic is the same. This is also why it is outside the loop--so that as long as one batch has data, we return a response.

I will add an else statement that continues the for loop so that we don't have a key error. Will also add a test for this.

datacommons/stat_vars.py
tjann: Let me know if this comment still holds after my response here: https://github.com/datacommonsorg/api-python/pull/155#discussion_r486546024

it would be a logic reversal with no behavioral changes.

shifucun: here i actually mean a wording change, use has_data as oppose to no_data for variable naming, just readability nit :)

tjann (commented):
PTAL. This shows the changes since the last review: https://github.com/datacommonsorg/api-python/pull/155/files/d24bf0597c8b3f2c857770ae04b083e75c4899d3..a8fad527090045e073589ec68c2fec26ca964505

tjann:
I was on the fence between doing lenient key error handling or throwing errors if REST response didn't follow the expectations.

The former might be nice for us to have more time to fix issues when REST changes, but also makes it difficult for us to know when something is wrong. So I opted for the latter in the spirit of failing fast.

beets:
The error handling here seems reasonable to me, since those are truly errors that should get looked at (and isn't a matter of a 1 missing data out of 1000 calls).

beets (approved):

shifucun (approved):
thanks for the update!

tjann:
Thanks all for the review!

PR #157, opened 2020-09-16
Make the batch size and number of batches also reflective of number o…

tjann:
…f stat_vars queried. Increase number of stat_vars in stress test. Tested by installing https://test.pypi.org/project/datacommons-pandas-tjann/0.0.12/ on the GH Colab, which now runs for the 3000 counties x 200 vars table builder.


shifucun (approved):
thanks for the updat!

PR #159, opened 2020-09-16
Remove the schema subdir from api-python

tjann:
The Data Commons schema is now being updated in https://github.com/datacommonsorg/data, and is proposed to being moved to a separate directory.

The code that hosted schema.datacommons.org has not been deployed for a while, and if we do revive schema.datacommons.org, the infrastructure will likely be redone.


beets:
Could you please add more detail to the PR description?

beets (approved):

PR #164, opened 2020-10-13
Handle statvar api no data.

tjann:


datacommons/stat_vars.py
```diff
@@ -77,7 +77,9 @@ def get_stat_value(place,
     try:
         res_json = utils._send_request(url, post=False, use_payload=False)
     except ValueError:
```
tjann, line 79: @shifucun we might want to also change REST behavior to not return 404

datacommons/examples/stat_vars.py
```diff
@@ -126,6 +126,13 @@ def call_str(pvs):
         dc.get_stat_all(["badPlaceId", "country/FRA"],
                         ["Median_Age_Person", "Count_Person"]))
 
+
+    print('\nWhen no data for get_stat_value')
+    pp.pprint(dc.get_stat_value('Lame', 'lame'))
```
beets, line 131: maybe choose another word for these examples?

beets (approved):
Yea, we should allow callers to differentiate between server error or missing data.

shifucun (approved):

PR #166, opened 2020-11-10
Release py1.4.3 and pd0.0.3 for returning empty data structures inste…

tjann:
…ad of erroring on no data.

Tested on TestPyPI. They are live on PyPI. Currently testing on tutorials and so far looking so good.


beets (approved):
thanks!

PR #175, opened 2021-10-10
fix flaky test on set_api_key_test.py::TestApiKey::test_query_no_api_key and set_api_key_test.py::TestApiKey::test_send_request_no_api_key

lonly7star:
This PR aims to fix the flaky test on **set_api_key_test.py::TestApiKey::test_query_no_api_key** and   **set_api_key_test.py::TestApiKey::test_send_request_no_api_key** so that the test could pass both individually and as part of the testing suit, also prevent testing fail in random order or repeat sequence.

#### The result

The test can pass when running the whole test suite, but fail when running the test suit multiple times or running the test individually. Both test method `test_query_no_api_key` and `test_send_request_no_api_key` has the same error.

#### Steps to reproduce the issue

1.  install pytest flaky finder with `pip install pytest-flakefinder`
2.  run pytest with flake-finder with command  `pytest -k . --flake-finder`
3.  or run pytest on its own `pytest set_api_key_test.py`

#### Issue of the code

The reason is the code tried to delete the existing os environmental key `utils._ENV_VAR_API_KEY`. However, the environmental variable is set by another test file` places_test.py`, so when directly runs the `set_api_key_test.py` the delete operation will fail due to the key not being found. `KeyError: 'DC_API_KEY'`

#### Proposed solution
Adding one line of if-statement checks if the key existing in the os environment. Delete if the key exists, skip the deletion if not. 
The same fixed is applied to both `test_query_no_api_key` and `test_send_request_no_api_key` method


lonly7star:
@googlebot I fixed it.

lonly7star:
@googlebot I fixed it.

lonly7star:
@googlebot I fixed it.

shifucun (approved):

beets:
Thanks for submitting a fix @lonly7star !

PR #176, opened 2022-03-01
Remove deprecated population and observation code

shifucun:
External users still try to access the get_observations API which has been deprecated a long time ago.

Later on we should consider make the code more Python 3 compatible (or deprecate Python 2 support) and update all documentations to reflect recent changes.


README.md
```diff
@@ -19,7 +18,7 @@ add the following line to your code:
 
     dc.set_api_key('YOUR-API-KEY')
```
beets, line 19: Should we just remove the note about the api key? We've removed it from the rest of our documentation (and the link on line 16 is a 404..)

beets (approved):
thanks for the update!

another thing we should look into is setting up continuous testing for our libraries so we can catch these earlier.

README.md
shifucun: Removed API key section

PR #177, opened 2022-03-10
Add superfund site notebooks

pradh:


sharadshriram (commented):
This looks good to me, Thanks Prashanth! 

One suggestion: At the end of [Accessing_Superfund_data_from_Data_Commons](https://github.com/pradh/api-python/blob/superfund1/notebooks/Accessing_Superfund_data_from_Data_Commons.ipynb), can we change the url to the `extended notebook` from [`colab.sandbox.google.com`](https://colab.sandbox.google.com/github/pradh/api-python/blob/superfund1/notebooks/Analyzing_SuperfundSites_with_Data_Commons.ipynb#scrollTo=Bdada1GF-VgS) to the  relative [Github url](notebooks/Analyzing_SuperfundSites_with_Data_Commons.ipynb). I think the `sandbox.google.com` domain is not accessible to the external users

pradh:
> The published notebooks will have URLs of this format (you can use this instead of your personal github):
> 
> https://colab.research.google.com/github/datacommonsorg/api-python/blob/master/notebooks/analyzing_obesity_prevalence.ipynb

Thanks.  Done.

PR #178, opened 2022-03-10
Fix various source notebook links

pradh:


beets (approved):
thanks for updating the rest too!

PR #182, opened 2022-10-06
[V2] Add node APIs and clean up packaging

shifucun:
- Add node.py with properties, property_values and triples API.
- Add key.py with API KEY util functions.
- Deprecate Python 2 support.
- Re-organize pip packing files .


datacommons/setup.py
```diff
@@ -23,15 +25,10 @@
 URL = 'https://github.com/datacommonsorg/api-python'
 EMAIL = 'support@datacommons.org'
 AUTHOR = 'datacommons.org'
-REQUIRES_PYTHON = '>=2.7'
+REQUIRES_PYTHON = '>=3.7'
```
juliawu, line 28: Do we know if anyone is still using python 2.7 and should be alerted about this change?

datacommons/node.py
```diff
@@ -0,0 +1,76 @@
...
+""" API to request node information.
+"""
+
+from typing import List
+
+from datacommons.requests import _post
+from datacommons.utils import _get_direction
+
+
+def properties(nodes: List[str], out: bool = True):
```
juliawu, line 23: Suggestion: include type hints for return values.

juliawu:
For my own understanding, is the intention for `property()`, `property_values()`, and `triples()` to supercede the current `get_property_labels()`, `get_property_values()`, and `get_triples()` while maintaining the same specs?

juliawu:
LGTM, leaving approval for @beets or @pradh who may have more detailed comments.

datacommons/setup.py
shifucun: The change will be pushed as a new pip version. So users with older version can still use python 2.7

shifucun:
> For my own understanding, is the intention for `property()`, `property_values()`, and `triples()` to supercede the current `get_property_labels()`, `get_property_values()`, and `get_triples()` while maintaining the same specs?

They will supercede the current APIs but the specs won't necessarily be maintained, especially for observation APIs coming later.

datacommons/node.py
shifucun: Done

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+""" API to request node information.
+"""
+
+from typing import Dict, List
+
+from datacommons.requests import _post
+from datacommons.utils import _get_direction
+
+
+def properties(nodes: List[str], out: bool = True) -> Dict[str, List[str]]:
```
beets, line 23: since it's a bool, perhaps `is_out` or `is_out_arc` for the second parameter

datacommons/key.py
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" API key related functions.
+"""
+
+import os
+
+# Environment variable for API key.
+_KEY_ENV = 'DC_API_KEY'
```
pradh, line 20: Does the user need to set it?  Should we talk about it in README/docs?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+    """Retrieves all the properties for a list of nodes.
+    Args:
+        nodes: Node DCIDs.
+        out: Whether this is out going property.
+    """
+    resp = _post(f'/v1/bulk/properties/{_get_direction(out)}', {
+        'nodes': nodes,
+    })
+    result = {}
+    for item in resp.get('data', []):
```
beets, line 33: should we iterate through the argument list of nodes instead to ensure the resulting object has all requested properties?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+
+from typing import Dict, List
+
+from datacommons.requests import _post
+from datacommons.utils import _get_direction
+
+
+def properties(nodes: List[str], out: bool = True) -> Dict[str, List[str]]:
+    """Retrieves all the properties for a list of nodes.
```
beets, line 24: I'm not as familiar with pypi - if the docstring gets converted into documentation, do we need to put a Returns here so users know what to expect?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+    result = {}
+    for item in resp.get('data', []):
+        node, properties = item['node'], item.get('properties', [])
+        result[node] = properties
+    return result
+
+
+def property_values(nodes: List[str],
+                    property: str,
+                    out: bool = True) -> Dict[str, List[str]]:
```
beets, line 41: same here wrt `out` (and below)

docs/release.md
```diff
@@ -0,0 +1,54 @@
+# Python API Release
```
pradh, line 1: Thanks for adding this!

datacommons/node.py
beets, line 24: Might be worth mentioning that we are not returning values, only available property keys?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+        result[node] = []
+        for v in values:
+            if 'dcid' in v:
+                result[node].append(v['dcid'])
+            else:
+                result[node].append(v['value'])
+    return result
+
+
+def triples(nodes: List[str],
```
pradh, line 64: should use the convention of <verb>_<thing>?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+    """Retrieves the triples for a node.
+    Args:
+        node: Node DCID.
+        out: Whether the returned property is out-going for the queried nodes.
+    """
+    resp = _post(f'/v1/bulk/triples/{_get_direction(out)}',
+                 data={'nodes': nodes})
+    result = {}
+    for item in resp.get('data', []):
+        node, triples = item['node'], item.get('triples', {})
```
pradh, line 75: I assume 'node' is always there?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+            else:
+                result[node].append(v['value'])
+    return result
+
+
+def triples(nodes: List[str],
+            out: bool = True) -> Dict[str, Dict[str, List[object]]]:
+    """Retrieves the triples for a node.
+    Args:
+        node: Node DCID.
```
pradh, line 68: nodes.  List of DCIDs?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+        out: Whether the returned property is out-going for the queried nodes.
+    """
```
pradh, line 70: Document return value?

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+
+def property_values(nodes: List[str],
+                    property: str,
+                    out: bool = True) -> Dict[str, List[str]]:
+    """Retrieves the property values for a list of nodes.
+    Args:
+        nodes: Node DCIDs.
+        property: The property label to query for.
+        out: Whether this is out going property.
+    """
```
pradh, line 47: Document return value/

datacommons/node.py
```diff
@@ -0,0 +1,79 @@
...
+
+from datacommons.requests import _post
+from datacommons.utils import _get_direction
+
+
+def properties(nodes: List[str], out: bool = True) -> Dict[str, List[str]]:
+    """Retrieves all the properties for a list of nodes.
+    Args:
+        nodes: Node DCIDs.
+        out: Whether this is out going property.
```
pradh, line 27: nit: Whether to return out going properties?

datacommons/requests.py
```diff
@@ -0,0 +1,41 @@
...
+    if api_key:
+        headers['x-api-key'] = api_key
+    try:
+        resp = requests.post(url, json=data, headers=headers)
+        if resp.status_code != 200:
+            raise Exception(
+                f'{resp.status_code}: {resp.reason}\n{resp.json()["message"]}')
+        return resp.json()
+    except requests.exceptions.Timeout:
+        raise Exception('data request timeout')
```
beets, line 39:
```suggestion
        raise Exception('Data request timed out, please try again.')
```

datacommons/node.py
beets: should we also add a try/catch for key errors so we can add a friendly error message

datacommons/setup.py
```diff
@@ -23,15 +25,10 @@
...
 VERSION = '1.4.3'
-
-REQUIRED = [
-    'six',
-]
-
+REQUIRED = ['six', 'requests']
```
beets, line 30: do we still need six if we're fully on py 3?

datacommons/test/node_test.py
```diff
@@ -0,0 +1,181 @@
...
+        response = datacommons.properties(["City"])
+        assert response == {
+            "City": ["name", "provenance", "subClassOf", "typeOf"]
+        }
+
+
+class TestPropertyValues(unittest.TestCase):
+
+    @mock.patch("datacommons.node._post")
+    def test_with_data(self, _post):
```
beets, line 50: can we also add a test with multiple values for a property?

pradh (commented):
Nice!

Main question about function name...

beets (approved):
thanks for updating the library!

datacommons/key.py
shifucun: Yes, the user will need to set it. We should add this back in docs.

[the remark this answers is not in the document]
shifucun: Yes, should pin our notebooks with the current version so they don't break.

datacommons/node.py
shifucun: Done

datacommons/node.py
shifucun: I think these docs are only meant for code, not for external documentations. I will put a short blurb of the return values

datacommons/node.py
shifucun: Done

datacommons/node.py
shifucun: Done

datacommons/node.py
shifucun: V1 REST will ensure all input have entries in the output, so here should be safe.

For error, can just let it throw the exception from inside _post(); it has added meaningful error msg there.

datacommons/node.py
shifucun: Done

datacommons/node.py
shifucun: Done

datacommons/node.py
shifucun: Yes, V1 always return the input node in each entry

datacommons/requests.py
shifucun: Done

datacommons/setup.py
shifucun: Will remove six after the old code is deleted.

datacommons/test/node_test.py
shifucun: Done

datacommons/node.py
shifucun: Done

datacommons/node.py
shifucun: Done

shifucun (commented):
thanks for review!

pradh (approved):

PR #185, opened 2022-11-04
Update Sparql query API to Python3

shifucun:


datacommons/sparql.py
```diff
@@ -0,0 +1,97 @@
...
+
+from datacommons.requests import _post
+
+
+def query(query_string, select=None):
+    """ Returns the results of executing a SPARQL query on the Data Commons graph.
+
+    Args:
+      query_string (:obj:`str`): The SPARQL query string.
+      select (:obj:`func` accepting a row in the query result): A function that
```
beets, line 27:
```suggestion
      select (:obj:`func` accepting a row of the query result): A function that
```

datacommons/sparql.py
```diff
@@ -0,0 +1,97 @@
...
+        selects rows to be returned by :code:`query`. This function accepts a row
+        in the results of executing :code:`query_string` and return True if and
```
beets, line 29:
```suggestion
        of the result of executing :code:`query_string` and returns True if and
```

beets (commented):
cool!

should we keep the "query" modules (rather than renaming to sparql) for consistency with the REST endpoint?

datacommons/sparql.py
shifucun: Done

datacommons/sparql.py
shifucun: Done

shifucun:
> cool!
> 
> should we keep the "query" modules (rather than renaming to sparql) for consistency with the REST endpoint?

Right now there is a module "query" which has a function "query". The library re-export the function in __init__.py:
`from datacommons.query import query`, which leaves "datacommons.query" ambiguous for test and import.

This is more of an internal change and users will still use "dc.query".

beets (approved):
ok makes sense. thanks bo

PR #186, opened 2022-11-07
Add bare bone APIs for observation and event

shifucun:
This PR aims to finalize the naming of Python API for observation and event.

### DIFF
--- a/datacommons/event.py
+++ b/datacommons/event.py
@@ -1,0 +1,22 @@
+# Copyright 2022 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" API to get event data."""
+
+
+def event_series_within_place():
+    return None
+
+
+def event_point_within_place():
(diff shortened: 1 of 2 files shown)


datacommons/event.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" API to get event data."""
+
+
+def event_series_within_place(event_type: str, containing_place: str,
+                              date: str):
```
pradh, line 18: We probably want to use variable here too?   Couple of options:

1. Use the same API as observation_series() and friends, except, depending on the variable (Count_EarthquakeEvent vs. List_EarthquakeEvent), the returned value will have entities per date, rather than a value per date.

2. Use very similar API structure as observation_series(), but these are "event" versions. We use the existing variables (e.g., `Count_EarthquakeEvent`), and this new API returns entities per data point.

datacommons/observation.py
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" API to get statistical observation data."""
+
+from typing import List
+
+
```
pradh, line 18: document return value?

pradh:
Do we want this as py-only API or REST API first?

datacommons/event.py
beets: my preference would be for (2)
rather than event, should we use `entity` to be more extensible?

PR #187, opened 2022-11-18
[Clean up] Use yapf to lint files; Remove dangling file links

shifucun:


beets (approved):
interesting that default yapf with google style uses tabstop=4, not 2

pradh:
> interesting that default yapf with google style uses tabstop=4, not 2

I've found our data repo setting at 4 slightly painful, but thought it was intentional.  I'd be all for switching to 2.

cloudbuild.yaml
```diff
@@ -4,4 +4,4 @@ steps:
     entrypoint: /bin/sh
     args:
       - -c
-      - "pip3 install -r requirements.txt && python3 -m pytest"
```
pradh, line 7: Didn't realize this repo used pytest.  Is it in some way better than unittest?

pradh (approved):

cloudbuild.yaml
beets: iirc from 2 years ago, it's better at handling module imports and discovery (we could say - find and run all files with *test.py from this dir). in the data repo though, we'd run into issues with how we're using FLAGS

datacommons/examples/places.py
```diff
@@ -24,50 +24,60 @@
...
+    print('Get place stats -- 2014')
+    stats = dc.get_stats(['geoId/05', 'geoId/06', 'dc/madDcid'],
+                         'dc/0hyp6tkn18vcb',
+                         obs_dates=['2014'])
+    print(stats)
 
-  print('Get place stats -- 2014 badly formatted')
-  stats = dc.get_stats(['geoId/05', 'geoId/06', 'dc/madDcid'], 'dc/0hyp6tkn18vcb', obs_dates='2014')
-  print(stats)
+    print('Get place stats -- 2014 badly formatted')
```
juliawu, line 61: Suggestion: specify how format is wrong. Perhaps "2014, badly formatted obs_dates"?

datacommons/places.py
```diff
@@ -26,225 +26,243 @@
...
+                if not time_series:
+                    continue
+                for date in list(time_series):
+                    if date not in obs_dates:
+                        time_series.pop(date)
+                res[geo] = stats
+    return res
+
+
+def get_related_places(dcids,
```
juliawu, line 197: This function is really cool!

juliawu (approved):
Focused my review on docstrings, LGTM!

cloudbuild.yaml
pradh: thanks.  both those are pretty painful today.  medium term wonder if it's worth switching data repo to pytest, or something better...

shifucun:
Updated indent to 2

cloudbuild.yaml
shifucun: Checking online, looks like pytest is more flexible, simple than unittest. It has many plugins which might be useful for the complicated case in data repo.

PR #196, opened 2023-06-21
Updating temperature modeling colab

jehangiramjad:


pradh (approved):
Could you perhaps summarize the changes in the description?  Looks like we might be adding new counties?

PR #198, opened 2024-04-04
Update intro to data commons notebook to fix covid 19 variable

juliawu:
Updates the COVID19 variable used in this notebook to `Count_MortalityEvent_COVID19`.  Other formatting changes are a result of re-exporting from colab, and colab's changes to default display settings.

Screenshot of last cell with the variable which got changed in red box:
image

You can also view a live version of the notebook being added here: https://colab.sandbox.google.com/github/juliawu/datacommons-api-python/blob/master/notebooks/intro_data_science/Data_Commons_For_Data_Science_Tutorial.ipynb


pradh (approved):
Hard to really review diffs, so blind LG :)

notebooks/intro_data_science/Data_Commons_For_Data_Science_Tutorial.ipynb
```diff
@@ -3,10 +3,9 @@
   "nbformat_minor": 0,
   "metadata": {
     "colab": {
-      "name": "Data Commons For Data Science Tutorial.ipynb",
```
jehangiramjad, line 6: why the removal of the name?

notebooks/intro_data_science/Data_Commons_For_Data_Science_Tutorial.ipynb
```diff
@@ -232,7 +231,7 @@
         "         \"geoId/065300\",   # Oakland\n",
         "         \"geoId/0644000\"]  # Los Angeles"
       ],
-      "execution_count": null,
+      "execution_count": 3,
```
jehangiramjad, line 234: Should the execution counts not be null? I guess it depends on whether we want the outputs to be visible by default or not?

jehangiramjad (approved):
Approving with a couple of questions. Why not also attach screenshots of the revised notebook so the byte encoded output can be seen in the review?

Thanks for the fix btw :)

notebooks/intro_data_science/Data_Commons_For_Data_Science_Tutorial.ipynb
juliawu: This was not a change I made intentionally, it looks like colab's export metadata no longer includes name compared to 2 years ago.

PR #203, opened 2024-12-04
Implement Data Classes to Model API Responses

jm-rivera:
**Note:**
Previous PR in the stack: #202
Next PR in the stack #204

---
This PR implements dataclasses to model API responses from all endpoints. These data classes represent various entities in the Data Commons knowledge graph. Dan's comments on the design document provide more details about the rationale.

### Data Classes

* [`datacommons_client/models/node.py`](diffhunk://#diff-1816caba4b54e536065474e7e040883c0e23d27651b707995781b8458066fd9fR1-R113): Added `Node`, `NodeGroup`, `Arcs`, and `Properties` data classes to represent nodes, groups of nodes, arcs, and properties in the Data Commons knowledge graph. 

* [`datacommons_client/models/observation.py`](diffhunk://#diff-f1f665d269a5d622ebcb7cc1f5f3fdaaa7556e4a29f6523f3e33bede0e5136b5R1-R147): Added `Observation`, `OrderedFacets`, `Variable`, and `Facet` data classes to represent observations, ordered facets, variables, and facets in the Data Commons knowledge graph.

* [`datacommons_client/models/resolve.py`](diffhunk://#diff-0ed570a995621dfe4a5781d43a21fe506a1fe141c35268c778c482f95fcd7fadR1-R69): Added `Candidate` and `Entity` data classes to represent candidates and entities in the resolution response. 

* [`datacommons_client/models/sparql.py`](diffhunk://#diff-8335343c9816ac0358beff5d328035ebafd60faf90fbc6606804c595f92f23dcR1-R47): Added `Cell` and `Row` data classes to represent cells and rows in a SPARQL query response. Note that the sparql endpoint does not (currently) work for custom instances.

### Unit Tests
In general, the tests for these classes are just making sure that the objects (especially objects that incorporate other objects) are built consistently and correctly. We avoid testing any basic python functionality.

* [`datacommons_client/tests/models/test_node_models.py`](diffhunk://#diff-4058485f3c30d988e24c32c4ac8f517a5ad4348f5989a14f62bb6c2b70b1504bR1-R89): Added unit tests for the `Node`, `NodeGroup`, `Arcs`, and `Properties` classes, testing their `from_json` methods with both complete and partial data.

[...]

PR #202, opened 2024-12-04
Add Initial Folder Structure for Data Commons Client

jm-rivera:
This is a housekeeping PR which adds the folder structure we plan to implement, based on the design document.
The folders are empty (other than `__init__` files).

We decided to put everything inside `/datacommons_client` while the DC team decides how they want to structure things with the different packages and libraries. The structure could always be refactored in the future, but this will allow us to develop in a relatively clean environment without deleting or moving things around.

PR #204, opened 2024-12-05
Introduce data classes to structure the responses from the API

jm-rivera:
**Note:**
~Previous PR in the stack: #203~ (merged)
Next PR in the stack: #205

---

This PR introduces data classes to handle structured responses from the Data Commons API. They encapsulate and build on the dataclasses that model the different objects present in the response.

Note that the branch will need to be rebased as other branches in the stack are merged.

New data classes for structured responses:

* [`DCResponse`](diffhunk://#diff-476a2040a5ddeb18723fb7f2f0c391e6881a357fcd8c723ad05c682b43e10572R1-R311): Represents a general structured response from the Data Commons API, including a method to get the next token for pagination.
* [`NodeResponse`](diffhunk://#diff-476a2040a5ddeb18723fb7f2f0c391e6881a357fcd8c723ad05c682b43e10572R1-R311): Handles the response for node data including methods to flatten properties and extracting arcs.
* [`ObservationResponse`](diffhunk://#diff-476a2040a5ddeb18723fb7f2f0c391e6881a357fcd8c723ad05c682b43e10572R1-R311): Manages observation data responses, with methods to convert observations into records.
* [`ResolveResponse`](diffhunk://#diff-476a2040a5ddeb18723fb7f2f0c391e6881a357fcd8c723ad05c682b43e10572R1-R311): Represents the response from the Resolve endpoint.
* [`SparqlResponse`](diffhunk://#diff-476a2040a5ddeb18723fb7f2f0c391e6881a357fcd8c723ad05c682b43e10572R1-R311): Represents the response from the Sparql endpoint.

This PR also includes tests for all the new classes and methods.


datacommons_client/utils/response.py
```diff
@@ -0,0 +1,311 @@
+from dataclasses import asdict, dataclass, field
```
dwnoble, line 1: What do you think about moving this file to the `datacommons_client/endpoints/` folder?

datacommons_client/utils/response.py
```diff
@@ -0,0 +1,311 @@
...
+        return f"<Raw Data Commons API response>"
+
+    @property
+    def next_token(self):
+        return self.json.get("nextToken")
+
+
+@dataclass
+class NodeResponse:
+    """Represents the response from the Data Commons API.
```
dwnoble, line 32: Can you note the specific endpoint that returns this response? (and for other *Response objects in this file)

datacommons_client/utils/response.py
jm-rivera: That probably makes the most sense indeed @dwnoble. Further down the stack, `response.py` contains specific Response classes for each of the endpoints. So keeping them together sounds good. I'll make the change.

jm-rivera: File move: https://github.com/datacommonsorg/api-python/pull/204/commits/f0b7757f03ce23e832263788e02f80d954117da3

datacommons_client/utils/response.py
jm-rivera: Improved docstrings: [b166513](https://github.com/datacommonsorg/api-python/pull/204/commits/b166513fa89d12012426457bfafeee649f31d930)

dwnoble (approved):
Thanks Jorge!

PR #205, opened 2024-12-06
Add custom error handling for API errors

jm-rivera:
**Note:**
~First PR in the stack: #202~ (merged)
~Second PR in the stack: #203~ (merged)
~Previous PR in the stack: #204~ (merged)
Next PR in the stack: #206

Will manually rebase as PRs in the stack are merged to make it easier to review the changes introduced in this PR.

---

This PR implements custom error handling for API errors. It also includes the related tests.
A bit unrelated but for housekeeping, the imports throughout the module are now passed through `isort` with Google profile for consistent ordering.

### Error Handling:
* [`datacommons_client/utils/error_hanlding.py`](diffhunk://#diff-ffb6d0b7dcf5bd8f5549995aefbce2c151c962a5dfa612eb39cef2ed231bf5ddR1-R78): Introduced new error handling classes (`DataCommonsError`, `APIError`, `DCConnectionError`, `DCStatusError`, `DCAuthenticationError`, `InvalidDCInstanceError`).
* [`datacommons_client/tests/utils/test_error_handling.py`](diffhunk://#diff-9ca5c8db3c443de222e50d1c0d3202e2647c0ec203e6003d9c02625f82ef028aR1-R68): Added tests for the different error classes.

PR #208, opened 2024-12-09
Add EndpointPayload ABC and Specific Payload Dataclasses for API Requests

jm-rivera:
**Note:**
~Second PR in the stack: #203~ (merged)
~Third PR in the stack: #204~ (merged)
~Fourth PR in the stack #205~ (merged)
~Previous PR in the stack #206~ (merged)
Next PR in the stack #209

Will manually rebase as PRs in the stack are merged to make it easier to review the changes introduced in this PR.

---

This PR adds 'payload' dataclasses for the API endpoints. 

The `EndpointPayload` class is an Abstract Base Class, which defines the structure for all other payload classes. Payload classes are in charge of normalizing and validating the payload before it is used for a POST request. They also implement a `.to_dict` property for easy access to the correctly structured payload data as a python dictionary.

### Details:

* [`datacommons_client/endpoints/payloads.py`](diffhunk://#diff-0df663029f2d5333d028b5a082a55786cf0958ef7c1bf1626b963c9117364714R1-R237): Added abstract base class `EndpointPayload` and specific payload classes (`NodePayload`, `ObservationPayload`, `ResolvePayload`, `SparqlPayload`) to structure, normalize, and validate payloads for different API requests. 

**Tests**

* [`datacommons_client/tests/endpoints/test_payloads.py`](diffhunk://#diff-b9daedced31aca35acf3db43e28cfd1bb864afdb540650c3e5b175a239733ad8R1-R134): Introduced test cases for each payload class.


datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+    @abstractmethod
+    def to_dict(self) -> dict:
+        """Convert the payload into a dictionary format for API requests."""
+        pass
+
+
+@dataclass
+class NodePayload(EndpointPayload):
+    """
+    A dataclass to structure, normalize, and validate the payload for a node API request.
```
dwnoble, line 36: Can you note that this is specifically for the /v2/node endpoint?

dwnoble: Same for the other payload below

dwnoble (commented):
Thanks @jm-rivera ! To make things a little simpler with this stack of PRs, it may be easier to just address all comments in this PR (208). what do you think?

datacommons_client/endpoints/payloads.py
jm-rivera: Done here: https://github.com/datacommonsorg/api-python/pull/208/commits/cda23078c42f3702295fdaff89ca445648f3cb4c

jm-rivera:
> Thanks @jm-rivera ! To make things a little simpler with this stack of PRs, it may be easier to just address all comments in this PR (208). what do you think?

Thanks for all your reviews @dwnoble!  To recap a bit where things stand now:
1.  #202 **(merged)**
2.  #203 **(merged)**

I have also rebased all of the PRs throughout the stack, resolving conflicts and incorporating the right changes. That should hopefully help a bit with the review. 

3.  #204 **new review requested**. I've rebased it and then made the changes you requested before requesting a new review.
4. #205: Also rebased. Contains all the changes from 204, plus I've fixed the filename typo. Awaiting a new review.
5. #206: Also rebased, incorporating all changes from 204 and 205. It's been approved but will wait until we merge 204 and 205 to rebase again and merge.

And #208 is here, I've rebased incorporating all changes from 204, 205, 206 and resolving conflicts.


--
Happy to follow whatever you prefer in terms of how/where we resolve anything that is outstanding.

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
+from abc import ABC
+from abc import abstractmethod
+from dataclasses import dataclass
+from dataclasses import field
+from enum import Enum
+from typing import Dict, Optional
+
+
+@dataclass
+class EndpointPayload(ABC):
```
keyurva, line 10: Are these intended to be request objects? If so, consider naming them with a `*Request` suffix.

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+    entity_expression: Optional[str] = None
+
+    def __post_init__(self):
+        """
+        Initializes the payload, performing validation and normalization.
+
+        Raises:
+            ValueError: If validation rules are violated.
+        """
+        self.RequiredSelect = {"variable", "entity"}
```
keyurva, line 101: Since variable and entity are always required, does it make sense to add them to `select` here if they are not set by the user?

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+            self.variable_dcids = [self.variable_dcids]
+
+        # Normalize entity
+        if isinstance(self.entity_dcids, str):
+            self.entity_dcids = [self.entity_dcids]
+
+        # Normalize date field
+        if self.date.upper() == "ALL":
+            self.date = ObservationDate.ALL
+        elif (self.date.upper() == "LATEST") or (self.date == ""):
```
keyurva, line 122: This makes latest the default. Is that the intent?

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        Validates the payload to ensure consistency and correctness.
+
+        Raises:
+            ValueError: If both `entity_dcids` and `entity_expression` are set,
+                        if neither is set, or if required fields are missing from `select`.
+        """
+
+        # Validate mutually exclusive entity fields
+        if bool(self.entity_dcids) + bool(self.entity_expression) != 1:
```
keyurva, line 135: nit: Can also be expressed as `bool(self.entity_dcids) == bool(self.entity_expression)` if that's easier to read.

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+        if not isinstance(self.expression, str):
+            raise ValueError("Expression must be a string.")
+
+    @property
+    def to_dict(self) -> dict:
+        return {"nodes": self.nodes, "property": self.expression}
+
+
+@dataclass
+class SparqlPayload(EndpointPayload):
```
keyurva, line 201: I was curious if you use the sparql endpoint? Asking because we've been discussing deprecating this endpoint. Let's not include it in this library if we can avoid it. (If you are using it then let's chat).

datacommons_client/endpoints/payloads.py
jm-rivera: As `EndpointRequestPayload`, `NodeRequestPayload`, etc? I think that would make things clearer indeed.

datacommons_client/endpoints/payloads.py
jm-rivera: Yes, I thought that could be a nice default. But would you prefer the default is not to filter the data in any way (so default to 'ALL')?

datacommons_client/endpoints/payloads.py
jm-rivera: We are not using it, so happy to remove it. I think you're probably right to deprecate it. The use cases seemed quite limited, and the rest of the API is intuitive so writing queries may not be the best way to get to the data

jm-rivera: We are not using it, so happy to remove it. I think you're probably right to deprecate it. The use cases seemed quite limited, and the rest of the API is intuitive so writing queries may not be the best way to get to the data

datacommons_client/endpoints/payloads.py
jm-rivera: resolved here: https://github.com/datacommonsorg/api-python/pull/208/commits/77386f410beef3e2f65ca07f617202617a403d0b

datacommons_client/endpoints/payloads.py
jm-rivera: Resolved here: https://github.com/datacommonsorg/api-python/pull/208/commits/ee19b36218b3b4a54611a75f59dd31ee675631fb

datacommons_client/endpoints/payloads.py
jm-rivera: So making them the default values? I think that would be convenient if you're happy with it being a default instead of just something we enforce. But the default could also be something slightly more practical (like directly defaulting to all `["date", "variable", "entity", "value"]` as most people will probably want the data, rather than just the variable-entity check

datacommons_client/endpoints/payloads.py
keyurva: Or just `EndpointRequest`, `NodeRequest`, etc. (Feel free to go with your preference)

datacommons_client/endpoints/payloads.py
keyurva: That works as well. Make sure to document it on the class though.

datacommons_client/endpoints/payloads.py
keyurva: @beets @dwnoble - any preference on what should be the default?

datacommons_client/endpoints/payloads.py
jm-rivera: Added the 4 select values as default here: https://github.com/datacommonsorg/api-python/pull/208/commits/6f19960347f05dc477c25f065c2b41adade73124

jm-rivera:
Hey @keyurva, I've just requested a new review. To summarise the changes since your last review:
- I addressed your comments. Only the question of the default for date is pending, but we could modify that at a later stage if the team needs more time to consider.
- I addressed Hanna's comments for the payload scripts, which they added on #209. I addressed on this branch to keep all changes to this module together.

keyurva (approved):
Thanks Jorge!

Yes, ok to change the default for date later if needed.

PR #209, opened 2024-12-09
 Implement `API` and `EndPoint` classes to interact with Data Commons API

jm-rivera:
**Note:**
~Second PR in the stack: #203~ (merged)
~Third PR in the stack: #204~ (merged)
~Fourth PR in the stack #205~ (merged)
~Fifth PR in the stack #206~ (merged)
~Previous PR in the stack: #208~ (merged)

This PR has been rebased to include all the changes in master

---

This PR introduces a new `API` class and `EndPoint` class. The `API` class handles environment setup and makes POST requests (i.e it is used to interface with the Data Commons API), while the `EndPoint` class represents specific endpoints (not yet implemented beyond this generic case) within the Data Commons API and uses the `API` instance to make requests.

* `datacommons_client/endpoints/base.py`: Added `API` class to handle environment setup, resolve base URLs, build headers, and make POST requests. Added `EndPoint` class to represent specific API endpoints and make POST requests using the `API` instance.

**Tests**
* `datacommons_client/tests/endpoints/test_base.py`: Added unit tests for the `API` and `EndPoint` classes.


datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+@dataclass
+class ObservationPayload(EndpointPayload):
+    """
+    A dataclass to structure, normalize, and validate the payload for an Observation V2 API request.
+
+    Attributes:
+        date (str): The date for which data is being requested.
+        variable_dcids (str | list[str]): One or more variable IDs for the data.
```
hqpho, line 82: For consistency with `nodes` in NodePayload, maybe we could call this `variables`? Alternatively we could change `nodes` to `node_dcids`?

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        self.RequiredSelect = {"variable", "entity"}
+        self.normalize()
+        self.validate()
+
+    def normalize(self):
+        """
+        Normalizes the payload for consistent internal representation.
+
+        - Converts `variable_dcids` and `entity_dcids` to lists if they are passed as strings.
```
hqpho, line 109: Add list item for date normalization?

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        Validates the payload to ensure consistency and correctness.
+
+        Raises:
+            ValueError: If both `entity_dcids` and `entity_expression` are set,
+                        if neither is set, or if required fields are missing from `select`.
+        """
+
+        # Validate mutually exclusive entity fields
+        if bool(self.entity_dcids) + bool(self.entity_expression) != 1:
```
hqpho, line 135: Optional: Could instead check if the two bools are equal (i.e. either both True or both False).

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+            raise ValueError(
+                "Exactly one of 'entity_dcids' or 'entity_expression' must be set."
+            )
+
+        # Check if all required fields are present
+        missing_fields = self.RequiredSelect - set(self.select)
+        if missing_fields:
+            raise ValueError(
+                f"'entity' and 'variable' must be selected (missing: {', '.join(missing_fields)})"
```
hqpho, line 144: Nit: Construct error message from self.RequiredSelect to future-proof it

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+            )
+
+        # Check all select fields are valid
+        [ObservationSelect(select_field) for select_field in self.select]
```
hqpho, line 148: What error results if `select_field` is not a key in `ObservationSelect`?

datacommons_client/endpoints/base.py
```diff
@@ -0,0 +1,132 @@
...
+        if not isinstance(payload, dict):
+            raise ValueError("Payload must be a dictionary.")
+
+        url = (
+            self.base_url if endpoint is None else f"{self.base_url}/{endpoint}"
+        )
+        return post_request(url=url, payload=payload, headers=self.headers)
+
+
+class EndPoint:
```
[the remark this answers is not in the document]
jm-rivera: Ah good catch! Yes happy to make the change throughout!

datacommons_client/endpoints/payloads.py
jm-rivera: Done here: https://github.com/datacommonsorg/api-python/pull/208/commits/77386f410beef3e2f65ca07f617202617a403d0b

But actually the payloads.py PR is #208 (there are other changes there you may want to check out)

datacommons_client/endpoints/payloads.py
jm-rivera: The `ObservationSelect` raises a `ValueError` in that case. But would you rather we raised a custom error?

Currently the error for <invalid_value> is:
`ValueError: '<invalid_value>' is not a valid ObservationSelect`

datacommons_client/endpoints/payloads.py
jm-rivera: Fixed here: https://github.com/datacommonsorg/api-python/pull/209/commits/ba827ddd79d44f7593115675b2505e2eca7a45fb

datacommons_client/endpoints/payloads.py
hqpho: I leave it up to you but my slight preference would be for a more descriptive error!

datacommons_client/endpoints/base.py
```diff
@@ -0,0 +1,132 @@
...
+        url: Optional[str] = None,
+    ):
+        """
+        Initializes the API instance.
+
+        Args:
+            api_key: The API key for authentication. Defaults to None.
+            dc_instance: The Data Commons instance domain. Ignored if `url` is provided.
+                         Defaults to 'datacommons.org' if both `url` and `dc_instance` are None.
+            url: A fully qualified URL for the base API. If provided, `dc_instance`
```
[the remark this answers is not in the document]
jm-rivera: I'll improve the docs to explain in more detail. The short answer here is that Dan wanted to have a way to specify the url exactly, in case this was being used for local development (or if the package was used as part of the website app). More context here: https://docs.google.com/document/d/1k2mwD6g1MYxOeZ7zlJlZg6fnUPvzMKsj8xXJBZNafYM/edit?disco=AAABZgvfLsY

datacommons_client/endpoints/payloads.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+@dataclass
+class NodePayload(EndpointPayload):
+    """
+    A dataclass to structure, normalize, and validate the payload for a Node V2 API request.
+
+    Attributes:
+        nodes (str | list[str]): The DCID(s) of the nodes to query.
+        prop (str): The property or relation expression(s) to query.
```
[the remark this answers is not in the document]
jm-rivera: It's a good point about consistency. I think it may be worth calling them both `expression` since both accept an expression. 

I originally had called it `prop` because the website code also uses that term (and the Resolve endpoint isn't used there, hence the difference in usuage).  

The only 'issue' with `expression` is that  the endpoints expect 'property', but I think `property` (or `prop`) is not a very clear term from the users' point of view (not least because it can take an expression, and in the case of Resolve it's not really expecting a 'property' at all )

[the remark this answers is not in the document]
hqpho: Personally I support an alias that is (a) consistent within this client library code and (b) increases clarity for users!

datacommons_client/endpoints/base.py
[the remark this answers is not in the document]
hqpho: Got it, thank you for the extra context!!

datacommons_client/endpoints/payloads.py
jm-rivera: Implemented here as a custom error: https://github.com/datacommonsorg/api-python/pull/208/commits/efe9e2340b0db99b22d6d0a0db4d98f8a1c19dcd

datacommons_client/endpoints/payloads.py
jm-rivera: Implemented here: https://github.com/datacommonsorg/api-python/pull/208/commits/acf4fbb7cc78e2683bec8d38b6f2bdf58da98957

datacommons_client/endpoints/payloads.py
jm-rivera: Yes to be clear about what we expect, I think the latter - renaming nodes to `node_dcids`. Implemented here: https://github.com/datacommonsorg/api-python/pull/208/commits/dfebec777ce9b083d71dc0e4b36a0353e3b9696c

datacommons_client/endpoints/base.py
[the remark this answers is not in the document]
jm-rivera: better docstrings here: https://github.com/datacommonsorg/api-python/pull/209/commits/e8575fedb5635b2eb17b73cf8923df99b99b8a68

datacommons_client/endpoints/base.py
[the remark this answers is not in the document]
jm-rivera: Renamed here: https://github.com/datacommonsorg/api-python/pull/209/commits/529790a4259cf693fa4343b9307d428d1a2e9f50

datacommons_client/endpoints/base.py
```diff
@@ -0,0 +1,132 @@
...
+            raise ValueError("Cannot provide both `dc_instance` and `url`.")
+
+        if not dc_instance and not url:
+            dc_instance = "datacommons.org"
+
+        self.headers = build_headers(api_key)
+
+        if url is not None:
+            # Use the given URL directly (strip trailing slash)
+            self.base_url = url.rstrip("/")
```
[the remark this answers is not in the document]
jm-rivera: Yes, very good point - I had missed it! Implemented here: https://github.com/datacommonsorg/api-python/pull/209/commits/bffcbc98cb9e1e82da04008897a2fa997020c3d0

jm-rivera:
Hi @hqpho, just requested a new review. Thank you for all the very useful feedback and comments. I addressed certain things via #208 (for the payloads file), and the rest here, but they are all reflected here after I rebased.

hqpho:
> Hi @hqpho, just requested a new review. Thank you for all the very useful feedback and comments. I addressed certain things via #208 (for the payloads file), and the rest here, but they are all reflected here after I rebased.

Looks great, thank you for the quick fixes!!

PR #210, opened 2025-01-18
Improve and fix tests

jm-rivera:
**Note:**
This is the first PR in the new stack.
First PR in the stack: #210
Next PR in the stack #211

Will manually rebase as PRs in the stack are merged to make it easier to review the changes introduced in this PR.

---

This PR fixes a few failing tests, which were not working properly after other refactoring. 
It also passes the tests code through isort and black for consistent formatting. 

FYI - somehow the check wasn't failing in previous commits. I think that it clears even if tests fail.

### DIFF
--- a/datacommons_client/tests/endpoints/test_base.py
+++ b/datacommons_client/tests/endpoints/test_base.py
@@ -1,47 +1,64 @@
 from unittest.mock import patch
 
-import pytest
-
 from datacommons_client.endpoints.base import API
 from datacommons_client.endpoints.base import Endpoint
+import pytest
 
 
-@patch("datacommons_client.endpoints.base.build_headers")
-@patch("datacommons_client.endpoints.base.resolve_instance_url")
+@patch(
+    "datacommons_client.endpoints.base.build_headers",
+    return_value={"Content-Type": "application/json"},
+)
+@patch(
+    "datacommons_client.endpoints.base.resolve_instance_url",
+    return_value="https://api.datacommons.org/v2",
+)
+@patch(
+    "datacommons_client.endpoints.base.check_instance_is_valid",
+    return_value="https://api.datacommons.org/v2",
+)
 def test_api_initialization_default(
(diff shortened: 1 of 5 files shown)


jm-rivera:
Hello @keyurva and @hqpho - I just requested a review from you, but please feel free to redirect as needed.

PR #210 is the first PR in this stack and it focuses only on fixing some failing tests. The next is #211 which also includes this code. As before I'll rebase things as they get approved to make things easier to review.

datacommons_client/tests/endpoints/test_base.py
```diff
@@ -128,10 +156,55 @@ def test_endpoint_post_request(mock_post_request):
...
         endpoint.post(payload=["invalid", "payload"])
+
+
+@patch(
+    "datacommons_client.endpoints.base.build_headers",
+    side_effect=lambda api_key: {"X-API-Key": api_key} if api_key else {},
+)
+@patch(
+    "datacommons_client.endpoints.base.check_instance_is_valid",
+    side_effect=lambda url: url.rstrip("/"),
```
hqpho, line 178: Why is this needed in the mock? As far as I can tell, `check_instance_is_valid` doesn't modify the URL.

hqpho (approved):
Thanks for flagging that test failures aren't blocking PRs! I put together #212 with a fix and will rebase on top of this once merged.

datacommons_client/tests/endpoints/test_base.py
jm-rivera: Thanks Hannah! You mean this particular side effect? You're right that the `.rstrip` is (now) done before calling `check_instance_is_valid`. I think I wrote the test before deciding that the check shouldn't do anything other than checking (and I moved the `rstrip` out to just before calling `check_instance_is_valid`). I'll remove the rstrip so that this test is just like the others (with `return_value` rather than `side_effect`)

PR #212, opened 2025-01-22
Improve pre-merge tests and code formatting

hqpho:
- Add `set -e` to the test script. This fixes an issue where PR checks have been passing even when tests fail because the test script overall exits with code 0.
- Add test script modes for env setup, lint fix, and running all tests. Before, "Python test" mode also set up the environment and ran formatting tests.
- Add isort alongside yapf. This follows precedent in the website repo.
- Add datacommons_client dir to the paths checked by yapf.

~This PR can't be merged until #210 (with Python test fixes) is merged. This PR will also need another commit with formatting fixes after rebasing.~ Rebased and ready!


jm-rivera:
Thanks @hqpho, this is really helpful to see! I realise we didn't fully land on the code style/formatting questions before we get started. I'd been running the code through `isort` and `black` (since we use black), but I see that you use `yapf`. 

From the files you modified, I gather that I should be formatting as:
```bash
isort . --profile google --line-length 80 && yapf -r -i -p --style='{based_on_style: google, indent_width: 2}' datacommons_client/
```

Does that look right? Should we be doing line-length 80, or some other? Perhaps that's already implied in the google profile/style for `isort` and `yapf`

hqpho:
> Thanks @hqpho, this is really helpful to see! I realise we didn't fully land on the code style/formatting questions before we get started. I'd been running the code through `isort` and `black` (since we use black), but I see that you use `yapf`. 
> 
> From the files you modified, I gather that I should be formatting as:
> ```bash
> isort . --profile google --line-length 80 && yapf -r -i -p --style='{based_on_style: google, indent_width: 2}' datacommons_client/
> ```
> 
> Does that look right? Should we be doing line-length 80, or some other? Perhaps that's already implied in the google profile/style for `isort` and `yapf`

Sorry for the disconnect, I didn't notice at first glance the subtle differences between run_tests.sh here and run_test.sh in the website repo! The isort and yapf arguments here are the same as the website repo now, though I don't know the history of choosing the settings there. Just aiming for consistency across Data Commons code.

Re line length for isort, it looks like the default is 79 and the Google profile doesn't override that. So you shouldn't need to pass that flag. Otherwise the formatting commands look good! Once this PR is merged you can alternatively run `./run_test.sh -f` from the repo root.

run_tests.sh
```diff
@@ -13,48 +13,92 @@
...
+function run_isort {
+  EXTRA_ARGS=$@
+  isort $EXTRA_ARGS --profile=google --skip-glob="$FORMAT_EXCLUDE_PATH" \
+    $FORMAT_INCLUDE_PATHS
+}
+
+function run_lint_test {
+  source .env/bin/activate
+  if ! run_yapf --diff; then
     echo "Fix lint errors by running: ./run_test.sh -f"
```
keyurva, line 51: Looks like it's called `run_test.sh` in the comments although the file is `run_tests.sh`. Let's either update the comments or change the file name to `run_test.sh` to be consistent with test files in other repos.

keyurva (approved):
Thanks!

PR #211, opened 2025-01-20
Add node-specific Endpoint classes

jm-rivera:
**Note:**
This is the first PR in the new stack.
First PR in the stack: #210

This is PR #211.

Will manually rebase as PRs in the stack are merged to make it easier to review the changes introduced in this PR.

---

This PR introduces several changes to the package. Broadly, it:
- Introduces the specific Endpoint classes and tests (details below)
- Does some housekeeping (fully removing the Sparql endpoint support, formatting code, cleaning some docstrings)
- Adds max_pages support to the Endpoints

### Endpoints:

* [`datacommons_client/endpoints/node.py`](diffhunk://#diff-eaab813e8b788ec171f0c4febfe77941b0a7cc1ee7d492460b10dd26a947dbecR1-R152)
* [`datacommons_client/endpoints/observation.py`](diffhunk://#diff-2f6f5222e02c9429c1e19de9eb99b3efaf5abe87755205cbcf10538551d321bfR1-R145)
* [`datacommons_client/endpoints/resolve.py`](diffhunk://#diff-07b83f8a4db05288c8eb68d75253ca986efd73f49449d71a9574a7332e743ee8R1-R172)

### Other changes

* [`datacommons_client/endpoints/base.py`](diffhunk://#diff-a91f2f9bd858e93e6faee9f4ef4c0957144f53f299d7f32bdebc1cc07c0d4395L64-R67): Updated `post` method to include an optional `max_pages` parameter for pagination support.
* [`datacommons_client/endpoints/payloads.py`](diffhunk://#diff-0df663029f2d5333d028b5a082a55786cf0958ef7c1bf1626b963c9117364714L102-R102): Modified `ObservationRequestPayload` to make the `select` field optional and added logic to set default values if not provided.
* [`datacommons_client/endpoints/response.py`](diffhunk://#diff-1a7afaaf02cb37fac9f1d5b74960bb43915fd9fd28cae214c2783294bd8af072L276-L314): Removed the `SparqlResponse` class to streamline the response handling.

PR #213, opened 2025-01-24
Add endpoint-specific Endpoint classes

jm-rivera:
**Note**: this was previously submitted as #211. I closed that one since the formatting changes had made it difficult to rebase and keep things relatively clean. This is basically the same code, but with a much cleaner Git history, and based off of the current `master` branch (with fixed tests and the expected formatting).

@keyurva and @hqpho - I just requested a review from you, but please feel free to redirect as needed.
Sorry in advance for the slightly bigger PR. Since the different Endpoint classes were just more specific implementations of the base Endpoint, I decided to put them all together instead of 3 separate PRs for quite similar behaviour.

---

This PR introduces several changes to the package. Broadly, it:
- Introduces the specific Endpoint classes and tests (details below)
- Removes the Sparql endpoint
- Adds max_pages support to the Endpoints

### Endpoints:

* [`datacommons_client/endpoints/node.py`](https://github.com/datacommonsorg/api-python/pull/213/files#diff-eaab813e8b788ec171f0c4febfe77941b0a7cc1ee7d492460b10dd26a947dbec)
* [`datacommons_client/endpoints/observation.py`](https://github.com/datacommonsorg/api-python/pull/213/files#diff-2f6f5222e02c9429c1e19de9eb99b3efaf5abe87755205cbcf10538551d321bf)
* [`datacommons_client/endpoints/resolve.py`](https://github.com/datacommonsorg/api-python/pull/213/files#diff-07b83f8a4db05288c8eb68d75253ca986efd73f49449d71a9574a7332e743ee8)

### Other changes

* [`datacommons_client/endpoints/base.py`](https://github.com/datacommonsorg/api-python/pull/213/files#diff-a91f2f9bd858e93e6faee9f4ef4c0957144f53f299d7f32bdebc1cc07c0d4395): Updated `post` method to include an optional `max_pages` parameter for pagination support.
* [`datacommons_client/endpoints/payloads.py`](https://github.com/datacommonsorg/api-python/pull/213/files#diff-0df663029f2d5333d028b5a082a55786cf0958ef7c1bf1626b963c9117364714): Modified `ObservationRequestPayload` to make the `select` field optional and added logic to set default values if not provided.


datacommons_client/endpoints/observation.py
```diff
@@ -0,0 +1,145 @@
...
+      select: Optional[list[ObservationSelect | str]] = None,
+      entity_dcids: Optional[str | list[str]] = None,
+      entity_expression: Optional[str] = None,
+  ) -> ObservationResponse:
+    """
+        Fetches data from the observation endpoint.
+
+        Args:
+            variable_dcids (str | list[str]): One or more variable IDs for the data.
+            date (str): The date for which data is being requested.
```
hqpho, line 37: Include `ObservationDate` in type?

datacommons_client/endpoints/observation.py
```diff
@@ -0,0 +1,145 @@
...
+        variable_dcids=variable_dcids,
+        select=select,
+        entity_dcids=entity_dcids,
+        entity_expression=entity_expression,
+    ).to_dict
+
+    # Send the request
+    return ObservationResponse.from_json(self.post(payload))
+
+  def fetch_latest_observation(
```
hqpho, line 59: Nit: `observation` -> `observations` for consistency with other methods

datacommons_client/endpoints/resolve.py
```diff
@@ -0,0 +1,170 @@
+from typing import Any, Optional
+
+from datacommons_client.endpoints.base import API
+from datacommons_client.endpoints.base import Endpoint
+from datacommons_client.endpoints.payloads import ResolveRequestPayload
+from datacommons_client.endpoints.response import ResolveResponse
+
+
+def flatten_resolve_response(data: ResolveResponse) -> dict[str, Any]:
```
hqpho, line 9: Is it possible to make the return type stricter? Maybe `dict[str, list[str] | str]` or is that not supported by Python?

datacommons_client/endpoints/resolve.py
```diff
@@ -0,0 +1,170 @@
...
+    node = entity.node
+    if len(entity.candidates) == 1:
+      items[node] = entity.candidates[0].dcid
+    else:
+      items[node] = [candidate.dcid for candidate in entity.candidates]
+
+  return items
+
+
+def resolve_correspondence_expression(from_type: str,
```
hqpho, line 32: It looks like this is meant to be a private helper for ResolveEndpoint. I'm new to the Python styleguide but I think the recommendation in this case is to append a single underscore to the start of the function name: https://google.github.io/styleguide/pyguide#3162-naming-conventions

datacommons_client/endpoints/resolve.py
```diff
@@ -0,0 +1,170 @@
...
+      node_dcids = [node_dcids]
+
+    # Construct the payload
+    payload = ResolveRequestPayload(node_dcids=node_dcids,
+                                    expression=expression).to_dict
+
+    # Send the request and return the response
+    return ResolveResponse.from_json(self.post(payload))
+
+  def fetch_dcid_by_name(self,
```
hqpho, line 86: Nit: `fetch_dcid` -> `fetch_dcids` here and elsewhere for consistency with other methods that can handle either one or multiple inputs.

datacommons_client/endpoints/resolve.py
```diff
@@ -0,0 +1,170 @@
...
+  def fetch_dcid_by_coordinates(
+      self,
+      latitude: str,
+      longitude: str,
+      entity_type: Optional[str] = None) -> ResolveResponse:
+    """
+        Fetches DCIDs for entities by their geographic coordinates.
+
+        Args:
+            latitude (str): Latitude of the entity.
```
hqpho, line 135: Optional: Include example values for lat and lon

datacommons_client/endpoints/resolve.py
```diff
@@ -0,0 +1,170 @@
...
+        Returns:
+            ResolveResponse: The response object containing the resolved DCIDs.
+        """
+    expression = resolve_correspondence_expression(from_type="geoCoordinate",
+                                                   to_type="dcid",
+                                                   entity_type=entity_type)
+    coordinates = f"{latitude}#{longitude}"
+    return self.fetch(node_dcids=coordinates, expression=expression)
+
+  def fetch_from_type_to_type(
```
hqpho, line 148: I see in the design this was called `get_entity_correspondence_dictionary`. WDYT of a name that splits the difference, maybe something like `fetch_entity_type_correspondence`?

hqpho (commented):
This is a partial review, I'll take a look at the test classes later today!

datacommons_client/tests/endpoints/test_observation_endpoint.py
```diff
@@ -0,0 +1,169 @@
+from unittest.mock import patch
+
+from datacommons_client.endpoints.base import API
+from datacommons_client.endpoints.observation import ObservationEndpoint
+from datacommons_client.endpoints.payloads import ObservationDate
+from datacommons_client.endpoints.response import ObservationResponse
+
+
+@patch(
```
hqpho, line 9: Is there a particular reason to use base function mocks here vs mock of API elsewhere? If not my preference would be to use one approach for all endpoint tests, with a slight preference for mocking API because it makes the tests more concise.

datacommons_client/tests/endpoints/test_base.py
```diff
@@ -82,12 +82,61 @@ def test_api_post_request(mock_check_instance, mock_post_request):
...
+  endpoint = Endpoint(endpoint="node", api=api)
+  payload = {"key": "value"}
+
+  response = endpoint.post(payload=payload, max_pages=5)
+  assert response == {"success": True}
+  mock_post_request.assert_called_once_with(
+      url="https://custom.api/v2/node",
+      payload=payload,
+      headers=api.headers,
+      max_pages=5,
```
hqpho, line 139: rm duplicate test?

datacommons_client/endpoints/base.py
```diff
@@ -52,81 +52,93 @@ def __init__(
...
+    """
     if not isinstance(payload, dict):
       raise ValueError("Payload must be a dictionary.")
 
     url = (self.base_url if endpoint is None else f"{self.base_url}/{endpoint}")
-    return post_request(url=url, payload=payload, headers=self.headers)
+    return post_request(url=url,
+                        payload=payload,
+                        headers=self.headers,
+                        max_pages=max_pages)
```
hqpho, line 92: When using `max_pages`, is `nextToken` still returned as part of the response? I wonder if we ought to strip it out when it can't be used via this client.

I'm curious too about the use case for `max_pages` and whether instead it might be preferable to still expose the ability to use the token. Or in any case if there should be some way to tell whether the response contains complete or truncated results (I suppose that's what leaving the token in the response accomplishes! Okay talking myself around to it...)

datacommons_client/tests/endpoints/test_observation_endpoint.py
```diff
@@ -0,0 +1,169 @@
...
+  api = API(url="https://custom.api/v2")
+  endpoint = ObservationEndpoint(api=api)
+
+  response = endpoint.fetch_latest_observations_by_entity(
+      variable_dcids="dc/VariableID",
+      entity_dcids=["dc/Entity1", "dc/Entity2"],
+  )
+
+  # Check the response
+  assert isinstance(response, ObservationResponse)
```
hqpho, line 110: Can we also assert on the contents of the response? Doesn't have to be in every test, we can have a dedicated test for it.

datacommons_client/endpoints/node.py
```diff
@@ -0,0 +1,145 @@
...
+            ```python
+            response = node_endpoint.fetch(
+                nodes=["geoId/06"],
+                property="<-"
+            )
+            print(response.data)
+            ```
+        """
+    # Normalize the input expression
+    expression = _normalize_expression_to_string(expression)
```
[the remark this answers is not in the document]
jm-rivera: For the `expression`, the payload currently only checks that it gets a string (and it currently only accepts a string). 

I currently normalize here to enable expressions such as:
`->[name, latitude, longitude]` which may contain a direction, multiple properties, and even a constraint. 

In the process of building them, the properties need to be listed as a comma-separated string, per [this bit of the docs](https://docs.datacommons.org/api/rest/v2/#relation-expressions:~:text=The%20following%20table%20describes%20symbols%20in%20the%20V2%20API%20relation%20expressions%3A).

datacommons_client/endpoints/node.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Construct the expression based on the direction and constraints.
+    if out:
+      expression = f"->{expression}"
+      if constraints:
+        expression += f"{{{constraints}}}"
+    else:
+      expression = f"<-{expression}"
+      if constraints:
+        expression += f"{{{constraints}}}"
```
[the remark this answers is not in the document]
jm-rivera: Yes, much better!

datacommons_client/endpoints/node.py
```diff
@@ -0,0 +1,145 @@
...
+from datacommons_client.endpoints.response import NodeResponse
+
+
+def _normalize_expression_to_string(expression: str | list[str]) -> str:
+  """Converts a list of expressions to a string."""
+  if isinstance(expression, str):
+    return expression
+
+  return (f"[{', '.join(expression)}]"
+          if isinstance(expression, list) else expression)
```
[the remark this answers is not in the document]
jm-rivera: Yes indeed!

datacommons_client/endpoints/node.py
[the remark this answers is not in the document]
jm-rivera: I moved the normalization function to the payload script and simplified where we're calling it.

datacommons_client/endpoints/node.py
```diff
@@ -0,0 +1,145 @@
...
+            out (bool): Whether to fetch outgoing properties. Defaults to True.
+
+        Returns:
+            NodeResponse: The response object containing the property values.
+
+        Example:
+            ```python
+            response = node_endpoint.fetch_property_values(
+                node_dcids=["geoId/06"],
+                expression="->name"
```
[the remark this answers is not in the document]
jm-rivera: I think in this particular case, `properties` is the better term. The `expression` is built from the direction, constraints, and properties, and it's the class that handles building it. I've updated the code to reflect that change (only where it expects properties to build the expression - but in `.fetch` I've kept the expression name, since it expects an expression rather than just the properties (even though the expression can be a single property, for exampel)

datacommons_client/endpoints/resolve.py
jm-rivera: It is supported! Changed.

datacommons_client/endpoints/resolve.py
jm-rivera: I changed the name because it was no longer a dictionary (originally we weren't modelling the API responses, now it returns a `ResolveResponse` (which can be used to produce the dictionary).

But I agree that your suggested name is a better option.

datacommons_client/tests/endpoints/test_observation_endpoint.py
jm-rivera: Not at all - you're right! I've simplified the observation and resolve tests to match using MagicMock for the API, like in node.

datacommons_client/tests/endpoints/test_base.py
jm-rivera: ooops

datacommons_client/tests/endpoints/test_observation_endpoint.py
jm-rivera: I'm not sure I follow - could you clarify the type of test you'd like to see?

I've written response [tests here](https://github.com/datacommonsorg/api-python/blob/747b18350770fe7ea3b7386a06e5252f3ad3a278/datacommons_client/tests/endpoints/test_response.py), which is why I only check the type. But happy to write additional tests if you see that tests for parts of the logic are missing.

datacommons_client/endpoints/base.py
jm-rivera: It's good that we discuss this in more detail. In short:
- `max_pages` is really only exposed to the user when they instantiate an `Endpoint` class which supports it. Right now that's only `NodeEndpoint`. It allows the user to limit the number of 'pages' fetched. If `None` then all are fetched.
- In its current implementation, it will fetch all pages up to `max_pages` or once no more pages are left (whichever happens first). 

The whole thing is sort of invisible to the user (in my thinking for convenience and simplicity). In practice, once the response object gets returned to the user (in the Node case `ObservationResponse`) the next_token is always `None`... which could be an argument to remove it, to avoid creating the impression that a full response was returned when it may have gotten truncated by `max_pages`. There isn't much utility for the `next_token` from any of these responses, since we don't have anything that allows the user to use it directly (other than a very low-level api call).

A couple of options:
- We could return the actual `next_token` as part of the Response object. In that case, `None` would mean the full content was returned and a token would mean that it was truncated by `max_pages`.
- Instead of returning a `next_token` (given the lack of practical use for it), we could add a flag like `truncated = True` to the Response object, to let the user know (very explicitly) that the response is not complete.

What do you think?

jm-rivera:
Thank you so much @hqpho, this was all very helpful! 

I've addressed everything, except for:
- [This comment](https://github.com/datacommonsorg/api-python/pull/213#discussion_r1929379005) where more details would be helpful
- [This comment](https://github.com/datacommonsorg/api-python/pull/213#discussion_r1929938324) where your views on the best way forward would be great.

datacommons_client/tests/endpoints/test_observation_endpoint.py
hqpho: Sure, it would just be a test to make sure that the data returned by post_request is actually contained in the response. It does overlap with the response tests, so you could argue we don't need it. Maybe it would just be good to have integration tests for a live request to each endpoint once this is all done.

datacommons_client/endpoints/base.py
hqpho: I think at this point my preference is to return all pages and not support `max_pages` at all (and also remove nextToken from the response). Then later if there is a need for this convenience layer to be customizable, we can decide which direction to go based on the actual usage. The reasoning for not adding it now is that it's trickier to take away functionality than to add it, so would prefer not to commit to anything new until we know for sure there is a need for it. Also open to your feedback on how you actually do plan to use this!

@keyurva Would you be able to weigh in for a second opinion? I know you had concerns about requests returning too much data.

keyurva: How about supporting both? i.e. fetch up to `max_pages` of content and if there is still more pages, return the resulting `next_token`. If the client wants to fetch more pages, they need to make an explicit call with the `next_token`.

This simplifies usage in most cases where the number of pages should not exceed `max_pages` while at the same time guarding against edge runaway cases.

Thoughts?

hqpho: Either that route (`max_pages` but `next_token` is supported to continue if necessary) or the route of removing `max_pages` for now with the option to support `next_token` with or without `max_pages` later would work for me. Up to you Jorge!

jm-rivera: Hi @hqpho @keyurva just a nudge on the above so I can push the last few updates to this PR. Let me know what you think!

hqpho (commented):
(formally bouncing back in case you were waiting on any more comments from me!)

hqpho (commented):
(formally bouncing back in case you were waiting on any more comments from me!)

jm-rivera:
> (formally bouncing back in case you were waiting on any more comments from me!)

Hey @hqpho just here please: https://github.com/datacommonsorg/api-python/pull/213#discussion_r1941438552

datacommons_client/endpoints/base.py
hqpho: Sorry to suggest yet another alternative, but what if instead we both (a) provide a pass-through to pagination (i.e. support `next_token` in the request), and (b) support a param `fetch_all_pages`? The same preview behavior you describe would be supported by setting `fetch_all_pages` to false. This would basically be optional auto pagination as described here: https://google.aip.dev/client-libraries/4233

If you feel strongly about not providing a pass-through to pagination, I suggest we leave the window open for it in the future by leaving `next_token` in the response and relying on that instead of a boolean `truncated` response field.

Sorry to be splitting hairs about this. I'm nervous about pagination patterns because the [Google API style guide](https://google.aip.dev/158) talks about pagination being difficult to change in the future.

jm-rivera: Thank you Hannah! I've implemented this last suggestion:
- `next_token` is returned as a string if the response is not final and `None` if the response is complete.
- `all_pages` is now an `Optional` parameter for all `NodeEndpoint` methods. To mirror the behaviour up to this point, defaults to `True`
- An `Optional` `next_token` string is accepted in the all `NodeEndpoint` methods.

And I've modified the tests (and added a couple) for this functionality.

datacommons_client/tests/endpoints/test_observation_endpoint.py
jm-rivera: I've made a note to come back to add integration tests as the last step!

hqpho (approved):

PR #216, opened 2025-03-03
Improve `url` handling

jm-rivera:
@kmoscoe identified an improvement we can make to how the `DataCommonsClient` class handles a fully resolved URL.

For context, during the design phase, @dwnoble requested having a way to provide a fully resolved URL, instead of the URL of Google's Data Commons (datacommons.org) or the URL of a custom Data Commons instance (like datacommons.one.org). 

That means that the `DataCommonsClient` class is initiated with:
```python
""" 
Args:
            api_key (Optional[str]): The API key for authentication. Defaults to None. Note that
                custom DC instances do not currently require an API key.
            dc_instance (Optional[str]): The Data Commons instance to use. Defaults to "datacommons.org".
            url (Optional[str]): A custom, fully resolved URL for the Data Commons API. Defaults to None.
"""
```

However, as @kmoscoe pointed out, if a `url` is used, it will return an error unless `dc_instance` is set to `None`. That's because both `dc_instance` and `url` cannot be set at the same time... and currently the default value of `dc_instance` is, for convenience, `datacommons.org`.

This PR introduces a small check (and it's associated test), to ignore `dc_instance` (if its default value is used) in cases when a `url` is provided.

That would allow for something like this to work:
```python
from datacommons_client.client import DataCommonsClient
client = DataCommonsClient(url="http://localhost:8080/core/api/v2/")
```
Instead of raising an error and requiring to also set `dc_instance` to `None`.

### DIFF
--- a/datacommons_client/client.py
+++ b/datacommons_client/client.py
@@ -45,6 +45,11 @@ class DataCommonsClient:
             dc_instance (Optional[str]): The Data Commons instance to use. Defaults to "datacommons.org".
             url (Optional[str]): A custom, fully resolved URL for the Data Commons API. Defaults to None.
         """
+    # If a fully resolved URL is provided, and the default dc_instance is used,
+    # ignore that default value
+    if dc_instance == "datacommons.org" and url:
+      dc_instance = None
+
     # Create an instance of the API class which will be injected to the endpoints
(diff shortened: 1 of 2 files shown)


kmoscoe (approved):
Thank you so much for this!

hqpho (approved):

PR #218, opened 2025-03-10
Fix some erroneous comments in node.py

kmoscoe:
There is no data() method on the NodeResponse.

### DIFF
--- a/datacommons_client/endpoints/node.py
+++ b/datacommons_client/endpoints/node.py
@@ -83,7 +83,7 @@ class NodeEndpoint(Endpoint):
 
         Example:
             ```python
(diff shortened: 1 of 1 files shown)


datacommons_client/endpoints/node.py
```diff
@@ -83,8 +83,8 @@ def fetch_property_labels(
 
         Example:
             ```python
-            response = node_endpoint.fetch_properties(node_dcids="geoId/06")
-            print(response.data)
+            response = node.fetch_labels(node_dcids="geoId/06")
```
kmoscoe, line 86:
```suggestion
            response = node.fetch_property_labels(node_dcids="geoId/06")
```

PR #220, opened 2025-03-14
Improve response output

jm-rivera:
This PR is in response to a discussion with @kmoscoe and @dwnoble regarding users access the REST API response as dictionaries/json strings.

To recap:
- We make POST requests to the relevant endpoint
- That returns a JSON string, which Python would parse as a dictionary. However, in order to provide strict typing and consistent structure, we transform the response into Python dataclasses.
- Before this PR, we would convert those dataclasses to Python dictionaries and provide that back as a `.json` property. The dictionary would be more verbose than the API response, as it would include null values (which the REST API response doesn't include)

This PR introduces a way to remove `None` values from serialised responses so they are more compact (and so that they mirror the REST API response). Additionally, it changes how the data is returned to improve clarity for users (by using clearer/more appropriate names):
- Remove the `.json` property. The rationale is that we weren't actually returning a json string but a dictionary. Also, given that it was a property, we couldn't take additional parameters to specify whether a 'compact' version should be returned.
- Add a `.to_dict()` method, which returns a dictionary. It takes a parameter `exclude_none` (which defaults to `True`) which excludes any keys which are None or empty lists*. I decided to call it `exclude_none` rather than `compact` to mirror how Pydantic deals with the same issue (model_dump optionally takes `exclude_none=True` in order to remove all `None` values). This prepares us for a future move to Pydantic (instead of Python dataclasses), as we've discussed.
- Add a `.to_json()` method, which returns a json string (via `json.dumps`). It takes the same `exclude_none` parameter described above.

@kmoscoe we would need to update the draft docs in order to call `.to_dict()` instead of `.json`. That would output Python dictionaries that look essentially the same as what would come out of the POST request to the REST API.

<br>

*This could have been a truthy/falsy check. That would omit and empty lists, empty strings, empty sets... but it would also exclude zero (which is falsy). Since zero doesn't equal missing in observations, we would want responses to include it. So instead, we should explicitly check for None and empty lists.

### DIFF
--- a/datacommons_client/endpoints/response.py
+++ b/datacommons_client/endpoints/response.py
@@ -1,6 +1,7 @@
 from dataclasses import asdict
 from dataclasses import dataclass
 from dataclasses import field
+import json
 from typing import Any, Dict, List
 
 from datacommons_client.models.node import Arcs
@@ -69,9 +70,29 @@ class NodeResponse:
   def get_properties(self) -> Dict:
     return flatten_properties(self.data)
 
-  @property
-  def json(self):
-    return asdict(self)
+  def to_dict(self, exclude_none: bool = True) -> Dict[str, Any]:
+    """Converts the NodeResponse instance to a dictionary.
+
+        Args:
+            exclude_none: If True, only include non-empty values in the response. Defaults to True.
+
+        Returns:
+            Dict[str, Any]: The dictionary representation of the NodeResponse instance
+        """
+    return _to_dict(self, exclude_none=exclude_none)
+
+  def to_json(self, exclude_none: bool = True) -> str:
+    """Converts the NodeResponse instance to a JSON string.
+
+        Args:
+            exclude_none: If True, only include non-empty values in the response. Defaults to True.
+
+        Returns:
+            str: The JSON string representation of the NodeResponse instance
+
+        """
+
+    return json.dumps(self.to_dict(exclude_none=exclude_none), indent=2)
 
 
 @dataclass
@@ -100,10 +121,6 @@ class ObservationResponse:
         },
     )
 
-  @property
-  def json(self):
-    return asdict(self)
-
   def get_data_by_entity(self) -> Dict:
     """Unpacks the data for each entity, for each variable.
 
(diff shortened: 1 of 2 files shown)


kmoscoe (approved):
This is totally awesome, Jorge!  I can't comment on the code, but the APIs are great. I will test it out and let you know if I have any feedback.

kmoscoe:
I've updated my (first) doc PR to cover these https://github.com/datacommonsorg/docsite/pull/576/files. Please re-review.
I'll work on updating the second PR as well.

datacommons_client/endpoints/response.py
```diff
@@ -69,9 +107,29 @@ def parse_data(data: Dict[str, Any]) -> Arcs | Properties:
   def get_properties(self) -> Dict:
     return flatten_properties(self.data)
 
-  @property
-  def json(self):
-    return asdict(self)
+  def to_dict(self, exclude_none: bool = True) -> Dict[str, Any]:
```
dwnoble, line 110: could to_dict and to_json be abstracted to a mixin class that the response classes inherit from?

datacommons_client/endpoints/response.py
```diff
@@ -14,7 +15,44 @@
...
+
+  def _remove_none(
+      data: List[Dict[str, Any]] | Dict[str, Any]) -> List | Dict[str, Any]:
+    """Recursively removes an empty value from a dictionary."""
+
+    if isinstance(data, dict):
+      return {
+          k: _remove_none(v)
+          for k, v in data.items()
+          if v is not None and v != []
```
dwnoble, line 42: Should we preserve empty lists (`[]`) to maintain the original API response shape? Or do these empty lists get created by the dataclass if no value is present for a list-type value?

datacommons_client/tests/endpoints/test_response.py
```diff
@@ -516,7 +591,114 @@ def test_resolve_response_json():
...
 
   # Assert that the resulting dictionary matches the original input
   assert result == input_data
+
+
+def test_resolve_response_dict_exclude_none():
+  """Test that ResolveResponse.to_dict and json are consistent."""
+  # Input dictionary
+  input_data = {
+      "entities": [
```
dwnoble, line 604: Nit: test adding an empty list [] as a value in the input data

dwnoble (approved):

datacommons_client/endpoints/response.py
jm-rivera: Absolutely - should have done it that way! I've refactored it accordingly

datacommons_client/endpoints/response.py
jm-rivera: Yes if it empty lists are returned by the REST API, we shoudl keep them. I've changed this check accordingly.

datacommons_client/tests/endpoints/test_response.py
jm-rivera: Added1

PR #225, opened 2025-03-20
feature: support property filters in `observations_dataframe` 

jm-rivera:
This is an exciting new feature, inspired by our recent work on supporting the Observation facet filters.

**TL;DR**
This PR makes it possible to specify property filters so that only the relevant observations are fetched. This could substantially reduce the amount of data transferred if users specify a particular `unit`, `measurementMethod`, `observationPeriod`, etc.

---

Sample usage:

```python

df = dc.observations_dataframe(
    variable_dcids=["Count_Person"],
    date="all",
    entity_dcids=["country/USA"],
    property_filters={
        "measurementMethod": ["CensusACS5yrSurvey", "OECDRegionalStatistics"]
    },
)
```

The above will make a first call to the Observations endpoint to get the facets information (without fetching the observations themselves). It will then find for which facets it should fetch observations, and make that more 'expensive' call only for the required facets.

In this case, 83 observations are returned. The `property_filters` parameter is optional.  Without filtering for "measurementMethod" it would return all measurementMethods, which total 507 observations.

The `property_filters` parameter takes in a dictionary. Users can specify any property as a key and a string or list of strings as a value.

---

To enable the behaviour described above, this PR introduces the following changes:
**DataCommonsClient** class
- A `property_filters` parameter for the `observations_dataframe` method.
- A helper function (`_find_filter_facet_ids` to find which facets should be fetched, based on the `property_filters`)

**ObservationsEndpoint** class
- Optional parameter to specify `select` in all convenience methods (instead of just for `fetch`)

**ObservationsSelect** enum
-  Support for "facet"

**ObservationResponse** class
- `get_facets_metadata` to extract metadata from the response object (including start date, end date, number of observations, properties like measurementMethod, units, etc)

The PR also includes new tests for this functionality and modifies some existing tests accordingly.

---

@kmoscoe FYI

### DIFF
--- a/datacommons_client/client.py
+++ b/datacommons_client/client.py
@@ -6,6 +6,7 @@ from datacommons_client.endpoints.observation import ObservationEndpoint
 from datacommons_client.endpoints.payloads import ObservationDate
 from datacommons_client.endpoints.resolve import ResolveEndpoint
 from datacommons_client.utils.decorators import requires_pandas
+from datacommons_client.utils.error_handling import NoDataForPropertyError
 
 try:
   import pandas as pd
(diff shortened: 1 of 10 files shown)


datacommons_client/client.py
```diff
@@ -58,6 +59,58 @@ def __init__(
...
 
+  def _find_filter_facet_ids(
+      self,
+      fetch_by: Literal["entity", "entity_type"],
+      date: ObservationDate | str,
+      variable_dcids: str | list[str],
+      entity_dcids: Literal["all"] | list[str] = "all",
+      entity_type: Optional[str] = None,
+      parent_entity: Optional[str] = None,
+      property_filters: Optional[dict] = None,
```
dwnoble, line 70: Nit: could you make this typing more precise: `Optional[dict[str, str | list[str]]]`

datacommons_client/endpoints/response.py
```diff
@@ -151,6 +151,65 @@ def get_observations_as_records(self) -> List[Dict[str, Any]]:
...
+                             value: str | list[str]) -> list[str]:
+    """Finds facet IDs that match a given property and value.
+
+        Args:
+            property_name (str): The property to match.
+            value (str | list): The value to match. Can be a string, number, or a list of values.
+        Returns:
+            list[str]: A list of facet IDs that match the property and value.
+        """
+    return [
```
dwnoble, line 206: Could you break this list comprehension up to make it more readable?

I know we haven't been strictly adhering to it, but here's a link to the google python style guide which has some rules of thumb to look at: https://google.github.io/styleguide/pyguide.html#274-decision

datacommons_client/client.py
```diff
@@ -66,6 +119,7 @@ def observations_dataframe(
       entity_dcids: Literal["all"] | list[str] = "all",
       entity_type: Optional[str] = None,
       parent_entity: Optional[str] = None,
+      property_filters: Optional[dict] = None,
```
dwnoble, line 122: Nit -update type hint

datacommons_client/client.py
```diff
@@ -74,15 +128,17 @@ def observations_dataframe(
...
-                Required if `entity_dcids="all"`. Defaults to None.
-            parent_entity (Optional[str], optional): The parent entity under which the target entities fall.
-                Used only when `entity_dcids="all"`. Defaults to None.
+        entity_dcids (Literal["all"] | list[str], optional): The entity DCIDs to retrieve data for.
+            Defaults to "all". DCIDs must include their type (e.g., "country/GTM" for Guatemala).
+        entity_type (Optional[str], optional): The type of entities to filter by when `entity_dcids="all"`.
+            Required if `entity_dcids="all"`. Defaults to None.
+        parent_entity (Optional[str], optional): The parent entity under which the target entities fall.
+            Used only when `entity_dcids="all"`. Defaults to None.
+        property_filters (Optional[dict], optional): An optional dictionary used to filter the data using
```
dwnoble, line 140: Could you clarify in the comment that this is for filtering by "observation" properties? I realize that is probably implied, but I think it can't hurt since this method also deals with entities and variables.

dwnoble (approved):
Thank Jorge! This looks like a really nice addition.

datacommons_client/endpoints/response.py
jm-rivera: Yes, very good point! I've refactored it. It's not the simplest loop but definitely a lot clearer now.

PR #227, opened 2025-03-28
`fetch_dcids_by_wikidata_id` improvement

jm-rivera:
With thanks to @kmoscoe for spotting the naming inconsistency, this PR:
- Renames `.fetch_dcid_by_wikidata_id` to `fetch_dcids_by_wikidata_id` to clarify lists are accepted.
- Adds a test for lists of `wikidata_ids`

### DIFF
--- a/datacommons_client/endpoints/resolve.py
+++ b/datacommons_client/endpoints/resolve.py
@@ -83,23 +83,23 @@ class ResolveEndpoint(Endpoint):
 
   def fetch_dcids_by_wikidata_id(
       self,
-      wikidata_id: str | list[str],
+      wikidata_ids: str | list[str],
       entity_type: Optional[str] = None) -> ResolveResponse:
     """
-        Fetches DCIDs for entities by their Wikidata IDs.
+          Fetches DCIDs for entities by their Wikidata IDs.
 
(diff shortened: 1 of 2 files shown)


datacommons_client/endpoints/resolve.py
```diff
@@ -105,7 +105,7 @@ def fetch_dcids_by_name(self,
 
     return self.fetch(node_ids=names, expression=expression)
 
-  def fetch_dcid_by_wikidata_id(
+  def fetch_dcids_by_wikidata_id(
       self,
       wikidata_id: str | list[str],
```
kmoscoe, line 110: Should this also be renamed to 'wikidata_ids'? I think it should, because elsewhere you have things like 'node_dcids' and 'node_ids' to indicate that it can be a list.

dwnoble: +1 to this

dwnoble (approved):
Thanks Jorge! Approved pending kara's comment.

datacommons_client/endpoints/resolve.py
jm-rivera: Done!

PR #228, opened 2025-03-28
Refactor `flatten_resolve_response`

jm-rivera:
Move the logic from a standalone `flatten_respolve_response` function, to a method of the Response object. That way it complements `to_dict`, `to_json` with `to_flat_dict`.

Example usage:

```python

response = dc.resolve.fetch_dcids_by_name(
    ["Guatemala", "France", "Germany", "United States", "United Kingdom"],
    entity_type="Country",
)
```

This would return a `ResolveResponse` object. 

Calling the regular `to_dict` would look like this, keeping the original response structure and nesting intact.

```python
response.to_dict()

{'entities': [{'candidates': [{'dcid': 'country/FRA'}, {'dcid': 'country/FXX'}],
               'node': 'France'},
              {'candidates': [{'dcid': 'country/DEU'}], 'node': 'Germany'},
              {'candidates': [{'dcid': 'country/GTM'}], 'node': 'Guatemala'},
              {'candidates': [{'dcid': 'country/GBR'}],
               'node': 'United Kingdom'},
              {'candidates': [{'dcid': 'country/USA'}],
               'node': 'United States'}]}
```

Calling `to_flat_dict` will create a direct mapping between the name and the dcid candidate(s).

```python
response.to_flat_dict()

{'France': ['country/FRA', 'country/FXX'],
 'Germany': 'country/DEU',
 'Guatemala': 'country/GTM',
 'United Kingdom': 'country/GBR',
 'United States': 'country/USA'}

```

Thanks to @kmoscoe for highlighting issues with clarity in the previous implementation!

### DIFF
--- a/datacommons_client/endpoints/resolve.py
+++ b/datacommons_client/endpoints/resolve.py
@@ -1,33 +1,9 @@
-from typing import Any, Optional
+from typing import Optional
 
 from datacommons_client.endpoints.base import API
 from datacommons_client.endpoints.base import Endpoint
 from datacommons_client.endpoints.payloads import ResolveRequestPayload
 from datacommons_client.endpoints.response import ResolveResponse
-
-
-def flatten_resolve_response(
-    data: ResolveResponse) -> dict[str, list[str] | str]:
-  """
-    Flattens resolved candidate data into a dictionary where each node maps to its candidates.
-
-    Args:
-        data (ResolveResponse): The response object containing the resolved data.
-
-    Returns:
-        dict[str, Any]: A dictionary mapping nodes to their candidates.
-        If a node has only one candidate, it maps directly to the candidate instead of a list.
-    """
-  items: dict[str, Any] = {}
-
-  for entity in data.entities:
-    node = entity.node
-    if len(entity.candidates) == 1:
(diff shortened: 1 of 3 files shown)


kmoscoe (commented):
I like the name "to_flat_dict". Would it make sense to rename the observations one "get_observations_as_record", so that we could present this as a more generic formatter? In fact, would it make sense to support for the node endpoint as well?

dwnoble (approved):

jm-rivera:
> I like the name "to_flat_dict". Would it make sense to rename the observations one "get_observations_as_record", so that we could present this as a more generic formatter? In fact, would it make sense to support for the node endpoint as well?

I'll have a look separately @kmoscoe. Still trying to figure out how to best implement some of your feedback on the .to_json for some of these methods, so I'll work on this there.

kmoscoe:
Hey Jorge -- just looking at this again: something like this would be very useful for some of the node methods. In fact, one of our tutorials includes a single method call which returns just the name-DCID mappings for places contained in another one. Would it be possible to implement this for all node methods too?
@clincoln8 who was wondering about this too

PR #229, opened 2025-03-31
Add `fetch_available_statistical_variables`

jm-rivera:
This PR is part of a group of PRs which will bring some key features from the Data Commons website API to the client library (e.g #230, #231).

**Fetch available statistical variables**: this is an implementation of [this feature of the DC website](https://github.com/datacommonsorg/website/blob/master/server/lib/fetch.py#L344-L363).

In short, this PR:
- Adds a `fetch_available_statistical_variables` method to the ObservationEndpoint. This method fetches all statvars which have observations for one or more entities.

It also includes tests for the method and a helper function which structures the data by entity : statvar_list.

Example usage:

```python
from datacommons_client import DataCommonsClient

dc = DataCommonsClient(dc_instance="datacommons.one.org")

variables = dc.observation.fetch_available_statistical_variables(
    entity_dcids=["africa", "country/TGO"]
)

{'africa': ['Area_FloodEvent',
            'eia/INTL.7-1-TJ.A',
            'Annual_Emissions_CarbonDioxideEquivalent100YearGlobalWarmingPotential_SteelManufacturing',
            'eia/INTL.35-12-QBTU.A',
            'sdg/EN_MAT_DOMCMPG.PRODUCT--MF421',
            'sdg/EN_MAT_DOMCMPG.PRODUCT--MF21',
            'sdg/SG_DMK_PARLCC_LC.AGE--Y0T45__SEX--M__PARLIAMENTARY_COMMITTEES--PC_DEFENCE',
            'eia/INTL.65-2-QBTU.A',
            'eia/INTL.57-1-MT.A',
            'eia/INTL.68-2-MT.A',
            ...
],
 'country/TGO': ['sdg/SG_GEN_EQPWN',
                 'Count_Person_25OrMoreYears_Female_MastersDegreeOrHigher_AsFractionOf_Count_Person_25OrMoreYears_Female',
                 'sdg/SE_ADT_ACTS.TYPE_OF_SKILL--SKILL_ICTPST',
                 'Amount_Debt_GBP_LenderArabBankforEconomicDevinAfrica_AsAFractionOf_Amount_Debt_LenderArabBankforEconomicDevinAfrica',
                 'worldBank/UIS_R_2_GPV_G2',
                 'worldBank/UIS_GTVP_2_GPV_F',
                 'DiffRelativeToAvg_1980_2010_MaxTemp_Daily_Hist_5PctProb_Greater_Atleast1DayADecade_CMIP6_Ensemble_SSP245',
[...]

PR #230, opened 2025-03-31
Add `fetch_entity_names` method

jm-rivera:
This PR is part of a group of PRs which will bring some key features from the Data Commons website API to the client library (e.g #229, #231)

**Fetch entity names**: this is an implementation of a few DC website features:
- The name endpoint of the [internal API](https://github.com/datacommonsorg/website/blob/master/server/routes/shared_api/place.py#L158-L171). `/api/place/name`
- The [internal code](https://github.com/datacommonsorg/website/blob/master/server/lib/shared.py#L22-L55) to fetch the english names from the `name` property.
- The [internal code](https://github.com/datacommonsorg/website/blob/master/server/routes/shared_api/place.py#L174-L213) to fetch the i18n name using the `nameWithLanguage` property (the website automatically resolves the user's locale, this method takes it as an argument)

In short, this PR:
- Adds a `fetch_entity_names` method to `NodeEndpoint`. This method takes one or more `entity_dcids` and fetches their name. It defaults to English, but other languages (`fr`, `es`, etc) can be selected using the optional `language` argument. Optionally, users can request to fallback to a particular language if the requested language is not available (to avoid sparse results in some languages).

Example usage:

```python

names = dc.node.fetch_entity_names(
    entity_dcids=[
        "africa",
        "country/GTM",
        "country/USA",
        "wikidataId/Q2608785",
    ],
    language="de",
)


{'africa': 'Afrika',
 'country/GTM': 'Guatemala',
 'country/USA': 'Vereinigte Staaten'}
```

And with the optional fallback to english:
```python
names = dc.node.fetch_entity_names(
    entity_dcids=[
        "africa",
        "country/GTM",
        "country/USA",
        "wikidataId/Q2608785",
    ],
    language="de",
    fallback_language="en",
)

{'africa': 'Afrika',
 'country/GTM': 'Guatemala',
 'country/USA': 'Vereinigte Staaten',
 'wikidataId/Q2608785': 'La Democracia'}

```

In the background, this method uses `name` or `nameWithLanguage` depending on what the user requests. `name` is much cheaper than `nameWithLanguage` since it only contains one value per entity. The method only requests and parses `nameWithLanguage` if a non-english language is selected.

A previous version used english as the fallback, available via a boolean. The latest version implements a more generic, Optional, `fallback_language` which takes a string (default to `None` for no fallback)

### DIFF
--- a/datacommons_client/endpoints/node.py
+++ b/datacommons_client/endpoints/node.py
@@ -196,7 +196,7 @@ class NodeEndpoint(Endpoint):
       entity_dcids: str | list[str],
       language: Optional[str] = DEFAULT_NAME_LANGUAGE,
       fallback_language: Optional[str] = None,
-  ) -> dict[str, str]:
+  ) -> dict[str, dict[str, str]]:
     """
         Fetches entity names in the specified language, with optional fallback to English.
         Args:
@@ -205,7 +205,8 @@ class NodeEndpoint(Endpoint):
           fallback_language: If provided, this language will be used as a fallback if the requested
             language is not available. If not provided, no fallback will be used.
         Returns:
-          A dictionary mapping each DCID to its name (in the requested or fallback language).
+          A dictionary mapping each DCID to a dictionary with the mapped name, language, and
+            the property used.
         """
 
     # Check if entity_dcids is a single string. If so, convert it to a list.
@@ -220,19 +221,24 @@ class NodeEndpoint(Endpoint):
     data = self.fetch_property_values(
         node_dcids=entity_dcids, properties=name_property).get_properties()
 
-    names: dict[str, str] = {}
+    names: dict[str, dict[str, str]] = {}
 
     # Iterate through the fetched data and populate the names dictionary.
     for dcid, properties in data.items():
       if language == "en":
         name = extract_name_from_english_name_property(properties=properties)
+        lang_used = "en"
       else:
-        name = extract_name_from_property_with_language(
+        name, lang_used = extract_name_from_property_with_language(
             properties=properties,
             language=language,
             fallback_language=fallback_language,
         )
       if name:
-        names[dcid] = name
+        names[dcid] = {
+            "value": name,
+            "language": lang_used,
+            "property": name_property,
+        }
 
     return names
--- a/datacommons_client/endpoints/node.py
+++ b/datacommons_client/endpoints/node.py
@@ -5,8 +5,11 @@ from datacommons_client.endpoints.base import Endpoint
 from datacommons_client.endpoints.payloads import NodeRequestPayload
 from datacommons_client.endpoints.payloads import normalize_properties_to_string
 from datacommons_client.endpoints.response import NodeResponse
+from datacommons_client.utils.names import DEFAULT_NAME_LANGUAGE
+from datacommons_client.utils.names import DEFAULT_NAME_PROPERTY
 from datacommons_client.utils.names import extract_name_from_english_name_property
 from datacommons_client.utils.names import extract_name_from_property_with_language
+from datacommons_client.utils.names import NAME_WITH_LANGUAGE_PROPERTY
 
 
 class NodeEndpoint(Endpoint):
@@ -191,16 +194,16 @@ class NodeEndpoint(Endpoint):
   def fetch_entity_names(
       self,
       entity_dcids: str | list[str],
-      language: Optional[str] = "en",
-      fallback_to_en: bool = False,
+      language: Optional[str] = DEFAULT_NAME_LANGUAGE,
+      fallback_language: Optional[str] = None,
   ) -> dict[str, str]:
     """
         Fetches entity names in the specified language, with optional fallback to English.
         Args:
           entity_dcids: A single DCID or a list of DCIDs to fetch names for.
-          language: Language code (e.g., "en", "es"). Defaults to "en".
-          fallback_to_en: If True, falls back to English if the desired language is not found.
-            Defaults to False.
+          language: Language code (e.g., "en", "es"). Defaults to "en" (DEFAULT_NAME_LANGUAGE).
+          fallback_language: If provided, this language will be used as a fallback if the requested
+            language is not available. If not provided, no fallback will be used.
         Returns:
           A dictionary mapping each DCID to its name (in the requested or fallback language).
         """
@@ -210,7 +213,8 @@ class NodeEndpoint(Endpoint):
       entity_dcids = [entity_dcids]
 
     # If langauge is English, use the more efficient 'name' property.
-    name_property = "name" if language == "en" else "nameWithLanguage"
+    name_property = (DEFAULT_NAME_PROPERTY if language == DEFAULT_NAME_LANGUAGE
+                     else NAME_WITH_LANGUAGE_PROPERTY)
 
     # Fetch names the given entity DCIDs.
     data = self.fetch_property_values(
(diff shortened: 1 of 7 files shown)


datacommons_client/endpoints/node.py
```diff
@@ -174,8 +180,55 @@ def fetch_all_classes(
...
+        Returns:
+          A dictionary mapping each DCID to its name (in the requested or fallback language).
+        """
+
+    # Check if entity_dcids is a single string. If so, convert it to a list.
+    if isinstance(entity_dcids, str):
+      entity_dcids = [entity_dcids]
+
+    # If langauge is English, use the more efficient 'name' property.
+    name_property = "name" if language == "en" else "nameWithLanguage"
```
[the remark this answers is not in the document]
jm-rivera: Done [here](https://github.com/datacommonsorg/api-python/pull/230/commits/8c27cf8131f0ca2dcb4009a92efccfe2dbff4930)

jm-rivera:
Thanks @dwnoble!
 
> * If i request a particular language and have english as a fallback, how do i tell from the response if i'm seeing the requested language or the fallback?
It's a very fair question. I don't have a simple suggestion... but can see two potential routes:
- It's on my list to implement logging. Right now everything happens a bit too silently for my liking. So a solution could be logging a info or warning when the fallback is used.
- we add an additional key to the response, which tracks any entities for which the fallback was used. That means we still keep a convenient mapping of entity_id -> name, but if the 'fallback' key is present (or if it has items, depending on how we implement), then the fallback was used for the countries contained in that list.

What do you think of either option? Or maybe you see another route?

> * Some smaller places may not have a "name" field today in the KG- example: https://datacommons.org/browser/wikidataId/Q4104190 . This is arguably a data issue - maybe in this example we should be setting "name" to the Bulgarian name, but the issue remains. I think on the website we return the DCID instead of the name, but im not sure if that's the best behavior.

In most of the REST API, if something doesn't exist it comes back empty. We don't currently infer or fill missing data on the client library side. I think I'd be in favour of keeping that behaviour consistent.

dwnoble:
> Thanks @dwnoble!
> 
> > * If i request a particular language and have english as a fallback, how do i tell from the response if i'm seeing the requested language or the fallback?
> 
> It's a very fair question. I don't have a simple suggestion... but can see two potential routes:
> 
> * It's on my list to implement logging. Right now everything happens a bit too silently for my liking. So a solution could be logging a info or warning when the fallback is used.
> * we add an additional key to the response, which tracks any entities for which the fallback was used. That means we still keep a convenient mapping of entity_id -> name, but if the 'fallback' key is present (or if it has items, depending on how we implement), then the fallback was used for the countries contained in that list.
> 
> What do you think of either option? Or maybe you see another route?
> 
> > * Some smaller places may not have a "name" field today in the KG- example: https://datacommons.org/browser/wikidataId/Q4104190 . This is arguably a data issue - maybe in this example we should be setting "name" to the Bulgarian name, but the issue remains. I think on the website we return the DCID instead of the name, but im not sure if that's the best behavior.
> 
> In most of the REST API, if something doesn't exist it comes back empty. We don't currently infer or fill missing data on the client library side. I think I'd be in favour of keeping that behaviour consistent.

Agreed that things happen a bit too silently right now, and that our website api currently makes some assumptions that are probably not obvious to users.  Extra logging sounds good, but I'd lean toward adding more metadata to the response.  What do you think about something like this?

```
{
  "africa": {
    "value": "Afrika",
    "property": "nameWithLanguage", // Could be 'name' if English was requested or used as fallback
    "language": "de" // The actual language of the value returned
  },
  "wikidataId/Q2608785": { // Example using fallback
     "value": "La Democracia", // The fallback name
     "property": "name", // Assuming 'name' is English and was the fallback
     // Optional: Explicitly note the request vs result?
     // "requestedLanguage": "de"
  }
  // ... other entities
}
```

dwnoble (commented):
> Thanks @dwnoble!
> 
> > * If i request a particular language and have english as a fallback, how do i tell from the response if i'm seeing the requested language or the fallback?
> 
> It's a very fair question. I don't have a simple suggestion... but can see two potential routes:
> 
> * It's on my list to implement logging. Right now everything happens a bit too silently for my liking. So a solution could be logging a info or warning when the fallback is used.
> * we add an additional key to the response, which tracks any entities for which the fallback was used. That means we still keep a convenient mapping of entity_id -> name, but if the 'fallback' key is present (or if it has items, depending on how we implement), then the fallback was used for the countries contained in that list.
> 
> What do you think of either option? Or maybe you see another route?
> 
> > * Some smaller places may not have a "name" field today in the KG- example: https://datacommons.org/browser/wikidataId/Q4104190 . This is arguably a data issue - maybe in this example we should be setting "name" to the Bulgarian name, but the issue remains. I think on the website we return the DCID instead of the name, but im not sure if that's the best behavior.
> 
> In most of the REST API, if something doesn't exist it comes back empty. We don't currently infer or fill missing data on the client library side. I think I'd be in favour of keeping that behaviour consistent.

Agreed that things happen a bit too silently right now, and that our website api currently makes some assumptions that are probably not obvious to users.  Extra logging sounds good, but I'd lean toward adding more metadata to the response.  What do you think about something like this?

```
{
  "africa": {
    "value": "Afrika",
    "property": "nameWithLanguage", // Could be 'name' if English was requested or used as fallback
    "language": "de" // The actual language of the value returned
  },
  "wikidataId/Q2608785": { // Example using fallback
     "value": "La Democracia", // The fallback name
     "property": "name", // Assuming 'name' is English and was the fallback
     // Optional: Explicitly note the request vs result?
     // "requestedLanguage": "de"
  }
  // ... other entities
}
```

datacommons_client/endpoints/node.py
```diff
@@ -220,19 +221,24 @@ def fetch_entity_names(
...
-        name = extract_name_from_property_with_language(
+        name, lang_used = extract_name_from_property_with_language(
             properties=properties,
             language=language,
             fallback_language=fallback_language,
         )
       if name:
-        names[dcid] = name
+        names[dcid] = {
+            "value": name,
```
dwnoble, line 239: Should this be a dataclass?

dwnoble:
> @dwnoble [Here's an implementation](https://github.com/datacommonsorg/api-python/pull/230/commits/bb8c0a3ce920b640714c829b3731e4f3c8457663) (with additional tests) for the type of return you suggested.
> 
> > Agreed that things happen a bit too silently right now, and that our website api currently makes some assumptions that are probably not obvious to users. Extra logging sounds good, but I'd lean toward adding more metadata to the response. What do you think about something like this?
> 
> It's definitely more explicit, and overall I think it's the right way to go. At first, I wanted something more directly usable for analysts who may want to quickly add names to the data they got. With this more complete dictionary, they would have to unpack `{k: v['value'] for k,v in result.items()}` to have a dictionary that directly takes the dcid and maps it to the name. It's a small inconvenience, but at least there would be no ambiguity.
> 
> Thanks for the suggestion, Dan!

It might en

dwnoble:
Accidentally hit close- apologies :)

datacommons_client/endpoints/node.py
jm-rivera: It did cross my mind. I didn't thinking of the unpacking (which would now be `{k: v.to_dict()['value'] for k,v in result.items()}`

For consistency, yes it should be. I've just[ implemented it here](https://github.com/datacommonsorg/api-python/pull/230/commits/39ce7f1302b7a6a59bdf8380de57bd3ee8a29ed8).

dwnoble (approved):

PR #234, opened 2025-04-08
Improvements suggested by kara

jm-rivera:
This PR addresses feedback from @kmoscoe:
- [Removes](77340ff98460f5ea101beeac545ec4bb7c696443) `fetch_latest_observations` as it didn’t really serve a useful role
- [Removes](5717ca261232eed78a3c49c5dd0ff8c033c16245)`fetch_latest_observations_by_entity` as it is basically the same as `fetch_observations_by_entity` but just with a default of the latest date.
- [Renames](f042f67e57d36b015be80a16c0e792f31fac92aa) `fetch_observations_by_entity` to `fetch_observations_by_entity_dcid`
- [Renames](bfa3f5978618ac845f10a84625cc12f90c42542e) `get_observations_as_records` to `to_observation_records`. This name more accurately reflects what the method does: it transforms nested observation data and facet metadata into a list of flat, record-like dictionaries suitable for analysis. This mirrors common data science terminology.

### DIFF
--- a/datacommons_client/client.py
+++ b/datacommons_client/client.py
@@ -87,12 +87,11 @@ class DataCommonsClient:
       return None
 
     if fetch_by == "entity":
-      observations = self.observation.fetch_observations_by_entity(
+      observations = self.observation.fetch_observations_by_entity_dcid(
           date=date,
           entity_dcids=entity_dcids,
           variable_dcids=variable_dcids,
-          select=["variable", "entity", "facet"],
-      )
+          select=["variable", "entity", "facet"])
     else:
       observations = self.observation.fetch_observations_by_entity_type(
           date=date,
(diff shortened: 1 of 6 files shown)


kmoscoe (approved):
Do you think it makes sense to have a fetch_observations_by_entity_expression? You can do this with fetch() so not sure it needs a convenience method. I'm fine either way.

dwnoble (approved):

dwnoble:
> Do you think it makes sense to have a fetch_observations_by_entity_expression? You can do this with fetch() so not sure it needs a convenience method. I'm fine either way.

I'll defer to you two if this would be helpful. I'd suggested we get the 2.0.0 release out, and can potentially add this down the road.

jm-rivera:
> Do you think it makes sense to have a fetch_observations_by_entity_expression? You can do this with fetch() so not sure it needs a convenience method. I'm fine either way.

> I'll defer to you two if this would be helpful. I'd suggested we get the 2.0.0 release out, and can potentially add this down the road.

Maybe we can explore for a `v2.1`. Writing expressions is quite advanced so maybe expecting users to just use fetch is fine. But happy to look again based on our usage once its out to see if we or others would find it helpful.

Thank you both!

PR #243, opened 2025-04-16
New Data Science tutorial

kmoscoe:
Create notebooks/v2 subdirectory
- Added an example of using the new fetch_statistical_variables method
- Edited and updated text


clincoln8 (approved):
Thanks Kara, looks good to me except for requesting this one change: 
https://screenshot.googleplex.com/BiZjxcKfYHBPWrT

PR #244, opened 2025-04-17
Remove auto-flattening for unpack_arcs.

clincoln8:
Currently the auto-flattening of unpack_arcs leads to the following scenario:

Two properties are requested for a given entity but data only exists for one of the properties. The REST api will not contain the property with no data in the response, therefore we cannot tell **from only the response** the difference between one property requested and one property out of multiple properties requested contains data.

An example:
```
node_resp = dc_client.node.fetch(node_dcids="bio/APOE", expression="<-[encodesGene,variantID]")
node_resp.get_properties()
> {'bio/APOE': [Node(dcid='bio/AB035149.1', name='AB035149.1', provenanceId='dc/base/NCBI_Gene'...), ...] ...}
```

In the response from get_properties(), we can't tell if it's for encodesGene or variantID.

A long term fix for this would be to store the original requested properties in the NodeResponse object and only do this "autoflattening" when there was one **requested** property. However, another consideration is that many return types for a single method can cause a lot confusion.

### DIFF
--- a/datacommons_client/tests/endpoints/test_response.py
+++ b/datacommons_client/tests/endpoints/test_response.py
@@ -182,7 +182,11 @@ def test_flatten_arcs():
   result = flatten_properties(response.data)
 
   assert "dc/03lw9rhpendw5" in result
-  assert result["dc/03lw9rhpendw5"].value == "191 Peachtree Tower"
+  assert result["dc/03lw9rhpendw5"] == {
+      "name": [
+          Node(value="191 Peachtree Tower", provenanceId="dc/base/EIA_860")
+      ]
+  }
 
 
 def test_flatten_multiple_arcs_with_multiple_nodes():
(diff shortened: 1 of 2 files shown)


jm-rivera (approved):
Thank you @clincoln8! You're absolutely right that return types should be consistent. In retrospect, I should err a bit less towards convenience and a bit more towards consistency.

jm-rivera:
> A long term fix for this would be to store the original requested properties in the NodeResponse object and only do this "autoflattening" when there was one requested property. However, another consideration is that many return types for a single method can cause a lot confusion.

I think we should just abandon auto-flattening entirely. This is more consistent and there may be other ways to make things easier for users downstream that don't involve quietly doing some magic in the background.

clincoln8:
Thanks both for the reviews!

PR #246, opened 2025-04-22
[NodeResponse] Add helpers to extract arcs from the response

clincoln8:
This adds two helper methods to simplify reading data from a NodeResponse.

`extract_connected_dcids`:
  * An example use case: Starting from a gene, I want to see associated diseases and indicators for that accociations. To accomplish this, I must find all of the dcids for the gene-disease-association nodes, then perform a follow up query for properties of those dcids. Using this helper method, it would look like:
    ```
    node_client = dc_client.node
    resp = node_client.fetch("bio/APOE", "<-geneID")
    gene_association_dcids = resp.extract_connected_dcids("bio/APOE", "geneID", connected_node_types="DiseaseGeneAssociation")
    association_metdata = node_client.fetch(gene_association_dcids, "->[relationshipAssociationType,relationshipEvidenceType, confidence]")
    ```
* The helper method makes it easy to extract just the dcids from the response that I care about without having to really dive into what the structure of the response is.

`extract_connected_nodes`: This has a similar use case to the first helper method, but returns the nodes for the arc that I'm interested in, instead of just the DCIDs.

---
This is a revised version of closed https://github.com/datacommonsorg/api-python/pull/238

@jm-rivera Let me know what you think! I can provide some more use cases if that would be helpful.

### DIFF
--- a/datacommons_client/endpoints/response.py
+++ b/datacommons_client/endpoints/response.py
@@ -1,12 +1,15 @@
 from dataclasses import dataclass
 from dataclasses import field
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from datacommons_client.models.base import SerializableMixin
 from datacommons_client.models.node import Arcs
 from datacommons_client.models.node import NextToken
+from datacommons_client.models.node import Node
 from datacommons_client.models.node import NodeDCID
+from datacommons_client.models.node import NodeGroup
 from datacommons_client.models.node import Properties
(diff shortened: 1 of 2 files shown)


kmoscoe (approved):
This looks great!  I think the names are fine.

dwnoble (approved):
Thanks @clincoln8!

jm-rivera (approved):
Thanks @clincoln8!

PR #250, opened 2025-04-25
Fix install command + minor edits

kmoscoe:
Fix install command to enable observations_dataframe method
Update Colab link
Various stylistic edits (heading levels, capitalization, etc.)


dwnoble (approved):
Thanks Kara!

PR #252, opened 2025-04-26
Fix bug in getting entity names

chejennifer:
Fixes this error in fetch_entity_names: https://screenshot.googleplex.com/Bm9BzSUnNYB87N9

There was a recent change that removes auto-flattening when unpacking arcs (https://github.com/datacommonsorg/api-python/pull/244), but fetch_entity_names was still assuming that values would be flattened.

Also updated the tests to use mock responses that are formatted the way the response would actually come back as.


clincoln8 (approved):
Thanks for the fix!!

To fix lint errors, run
```
pip install hatch
hatch env create
hatch run lint:format
``` 
(https://github.com/datacommonsorg/api-python/blob/master/docs/development.md#code-style-and-linting)
this is different than the styling used in website repo.

jm-rivera (approved):
Thank you for catching this and improving the tests @chejennifer!

PR #262, opened 2025-05-09
Fix broken feature engineering colab

kmoscoe:


kmoscoe:
This PR fixes the Data Science lesson on feature engineering, which was based on a statvar and places for which there isn't data anymore. It replaces that statvar with another one. 
Also fixes a couple of bugs (incorrectly named variables) that caused execution to fail.

clincoln8 (approved):
Thanks for the fixes!!

PR #287, opened 2026-01-23
Remove colabs that use legacy python client

n-h-diaz:
The packages have been deprecated, so cleaning up old python colabs.

This also restructures the folders to remove the nested v2 folder.


kmoscoe (approved):
You'll have to make some updates to the docsite as well to change the links to these files: https://docs.datacommons.org/api/python/v2/tutorials.html and https://docs.datacommons.org/courseware/intro_data_science.html

PR #288, opened 2026-01-23
Remove obsolete code for deprecated python client

n-h-diaz:
This PR removes code for the deprecated datacommons and datacommons_pandas packages

It also adds a top level README that points to the datacommons-client README and updates the deprecation notice and removes release instructions for the deprecated packages


beets (approved):
ty for the cleanup natalie!

